PR #7, opened 2022-07-27
Search Improvements

blattm:
This PR:
- makes search more robust by using get params
- splits the search api to allow searching for functions, samples, families and pichashes separately
- enables cursor based scrolling through search results

However, it:
- breaks the search of MemoryStorage
- does not address performance issues of filtering data within python instead of MongoDB


danielplohmann (approved):

danielplohmann:
thank you!

PR #9, opened 2022-07-28
fast tokenized search for MongoDbStorage

blattm:
Changes:
- use regex in MongoDB instead of python's `in` operation
- make search case insensitive
- tokenize search term by splitting it at `' '` characters. Every token needs to be found.


danielplohmann (approved):
thx!

PR #12, opened 2022-08-05
Fix family search

blattm:
Fixes the id match of family search, which was broken when FamilyEntries were introduced.


danielplohmann:
thx!

PR #36, opened 2023-08-23
Fix: remove redundant isfile call

yankovs:


danielplohmann (approved):
looks like I was sleeping while coding that lol. thx!

PR #41, opened 2023-09-13
Fix: Unify requitements in setup.py with reqirements.txt

yankovs:
Not to have duplicate requirements. Makes maintenance less error-prone.


danielplohmann:
Thanks, I only added an exclusion for gunicorn when on Windows with an addition to the requirements file.

PR #63, opened 2024-01-31
non-pool processing now matches new calculation_function tuple return…

dannyquist:
Copies the new format for calculation_function and matching the pool processing code. 

When running with `MINHASH_POOL_MATCHING=False` the results would not be returned in the correct format, resulting in an exception later in the code.


danielplohmann (approved):

danielplohmann:
Thanks for this extension, I must have missed to adapt that part - also I'm always running the parallel execution so I had no chance to notice it. :D

PR #70, opened 2024-03-28
Delete failed jobs on db cleanup

yankovs:


yankovs:
Looking the the failed test, I may be wrong but it looks like the fails is unrelated to the changes here. Regression issue with SMDA changes?

danielplohmann:
> Looking the the failed test, I may be wrong but it looks like the fail is unrelated to the changes here. Regression issue with SMDA changes?

yeah, that indeed looks like SMDA issues only. I'll fix that while integrating your PR. :)

danielplohmann:
yep, test ran through now, excellent. :)

PR #77, opened 2024-06-10
Fix: make job deletion safer

### LINKED ISSUE #73: Binary removal on job deletion might cause issues for certain files
Hey!

Some thoughts on a scenario: let's say sample `xyz.exe` is indexed in an MCRIT instance, i.e., added using an `addBinarySample` job. Later, let's assume the same `xyz.exe` sample was queried, for example using `getMatchesForUnmappedBinary`.

From what I saw, I guess for optimizations sake, files in `fs.files` can have a list of jobs, meaning that a file is saved once across multiple jobs. The issue is that from what I saw experimenting with this, both the job types will be stored in the `fs.files` entry for this file.

Then, when the `addBinarySample` is a week old (for example), it will be deleted during cleanup. Not sure what happens then when trying to use the system.

~Solution I thought about: only delete a sample from gridfs when (1) it's not indexed, AND, (2) all of the referenced jobs for sample are >= TLL (user defined) old, no new job for it.~

edit: I think that the existing `clean` method will work for this, it deletes the job_id from the `fs.files` entry which should mitigate the issue. Delete job_ids from the list of the file in `fs.files`, only delete the file if is it safe to delete - if the list of jobs is empty.

yankovs:
Fixes #73 

Helps with the case where the binary was both indexed and queried, or the same binary queried multiple times in which case we wouldn't want to remove it unless all jobs are deleted

### DIFF
--- a/mcrit/libs/mongoqueue.py
+++ b/mcrit/libs/mongoqueue.py
@@ -377,7 +377,7 @@ class MongoQueue(object):
                     if self._getFs().count_documents(
                         {"_id": file_object_id, "metadata.jobs": [], "metadata.tmp_lock": 0}
                     ) > 0:
-                        self._getFs().delete(ObjectId(file_object_id))
+                        self._getFs().delete(file_object_id)
             if with_result:
                 # delete result from GridFS  
                 self._getFs().delete(ObjectId(deletable_job["result"]))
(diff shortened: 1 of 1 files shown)


danielplohmann:
I totally agree, this is one quick and easy improvement, thanks for adding this!

PR #78, opened 2024-06-10
Log from worker subprocess, check if result_id is received

### LINKED ISSUE #76: Spawningworker: logging, check if process crashed
Hey!

I've tried the new spawning worker type, and I have some remarks about it:

In this section:
https://github.com/danielplohmann/mcrit/blob/90d3344700842f63e15a9d7e91e8d5ce99831b5d/mcrit/SpawningWorker.py#L90-L105
it should be taken into account that the singlejobworker subprocess might crash. We've noticed cases where it does crash, and in those cases what happens is that `result_id = None` is returned, and `Finished Remote Job with resuld_id: None` is logged. This causes the job to appear as finished in the web, but trying to access the result will result in an error page saying the result doesn't exist.

I think this can be easily handled with (plus maybe logic to mark job as failed):
```diff
with job as j:
                LOGGER.info("Processing Remote Job: %s", job)
                result_id = self._executeJobPayload(j["payload"], job)
-               # result should have already been persisted by the child process,we repeat it here to close the job for the queue
-               job.result = result_id
-               LOGGER.info("Finished Remote Job with result_id: %s", result_id)
+              if result_id:
+                  # result should have already been persisted by the child process,we repeat it here to close the job for the queue
+                  job.result = result_id
+                  LOGGER.info("Finished Remote Job with result_id: %s", result_id)
```

One other thing I wanted to mention, is that logs from the child singlejobworker process aren't appearing in logs when for example doing `docker logs <worker-container-id>`. Since you already get stdout and stderr of the child process, I think it's just a matter of logging them from the spawningworker parent process

Other than this the feature seems to be working great, and thank you for taking the time to implement it :)

yankovs:
Fixes #76 

Should anything be done on the else branch in line 127? Maybe marking job as failed?

### DIFF
--- a/mcrit/SpawningWorker.py
+++ b/mcrit/SpawningWorker.py
@@ -117,15 +117,14 @@ class SpawningWorker(Worker):
             self.queue.clean()
             self.t_last_cleanup = time.time()
         try:
-            with job as j:
-                LOGGER.info("Processing Remote Job: %s", job)
-                result_id = self._executeJobPayload(j["payload"], job)
-                if result_id:
-                    # result should have already been persisted by the child process,we repeat it here to close the job for the queue
-                    job.result = result_id
-                    LOGGER.info("Finished Remote Job with result_id: %s", result_id)
-                else:
-                    LOGGER.info("Failed Running Remote Job: %s", job)
+            LOGGER.info("Processing Remote Job: %s", job)
+            result_id = self._executeJobPayload(j["payload"], job)
+            if result_id:
+                # result should have already been persisted by the child process,we repeat it here to close the job for the queue
+                job.result = result_id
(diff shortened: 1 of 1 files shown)


danielplohmann:
Hey, I have a quick question: 
Why is it needed to remove the context manager for the jobs? 
Did that part break something or cause issues?
I'm asking because when jobs exit their context, they [do some stuff](https://github.com/danielplohmann/mcrit/blob/2bbe7f5310fbabeea002d42b0581f5612c483f8a/mcrit/libs/mongoqueue.py#L715), especially the `complete()` updates the [queue collection counters](https://github.com/danielplohmann/mcrit/blob/2bbe7f5310fbabeea002d42b0581f5612c483f8a/mcrit/libs/mongoqueue.py#L658).

PR #81, opened 2025-04-22
minor improvements

r0ny123:


danielplohmann (approved):
thanks, I always forget that it's possible to define module entry points for nicer CLI syntax. :)

PR #100, opened 2026-08-04
Fix pichash encoding in MongoDB search conditions

r0ny123:
### Motivation

Resolves the `# TODO: fix` in `MongoSearchTranspiler.visitSearchConditionNode()`:

```python
# TODO: fix
MongoDbStorage._encodePichash(None, condition)
```

Two problems with that call:

1. It was invoked on *every* search condition with `None` as `self`, which only worked because `_encodePichash()` happens to not use the instance.
2. For any operator other than `=`, the condition value is an operator dict (e.g. `{"$ne": 0x1234}`), and `hex()` on a dict raises `TypeError`. So a search such as `pichash!=0x1234` crashed instead of returning results.

### Changes

* Only call `_encodePichash()` when the condition is actually about `pichash`.
* Handle operator dicts in `_encodePichash()` / `_decodePichash()`, encoding the inner operand.
* Reject range operators (`<`, `<=`, `>`, `>=`) on `pichash` with a `ValueError`. `_pichash` is stored as a variable-length `hex()` string, so a Mongo range comparison on it is lexicographic and would return wrong results — better to say so than to answer incorrectly.
* Make the `_encode*` / `_decode*` conversion helpers `@staticmethod`, which is what the transpiler needed in the first place.
* Add `tests/testMongoSearchTranspiler.py` for the encoded/rejected cases.

### Note

I deliberately kept the stored encoding as `hex()`. Making pichash range queries actually work would mean switching `_pichash` to a fixed-width, zero-padded representation (`format(v, "016x")`), which is a breaking change for existing databases and would also make `_pichash` inconsistent with the `hash` entries inside `_picblockhashes`. That seems like a migration you'd want to decide on separately.

### DIFF
--- a/mcrit/storage/MongoDbStorage.py
+++ b/mcrit/storage/MongoDbStorage.py
@@ -104,12 +104,15 @@ class MongoSearchTranspiler(BaseVisitor):
             except Exception:
                 pass
         mongo_operator = operator_to_mongo[node.operator]
+        if node.field == "pichash" and mongo_operator in ("$lt", "$lte", "$gt", "$gte"):
+            # pichashes are stored as hex strings, on which range comparisons are not meaningful
+            raise ValueError("Range operators are not supported for the field 'pichash'.")
         if mongo_operator is None:
             condition = {node.field: value}
         else:
             condition = {node.field: {mongo_operator: value}}
-        # TODO: fix
-        MongoDbStorage._encodePichash(None, condition)
+        if node.field == "pichash":
+            MongoDbStorage._encodePichash(condition)
         return condition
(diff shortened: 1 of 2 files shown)


danielplohmann:
Landed in main as ab41635 (rebased onto the post-1.7.0 storage changes - the conflict was just the xcfg-split helper block that now sits between `_decodeXcfg` and `_encodePichash`; the fix itself applied unchanged). Full suite green at 164 tests including the new transpiler cases. Thanks for this - the `_encodePichash(None, condition)` call had been waiting for exactly this treatment.

PR #116, opened 2026-08-10
Use np.count_nonzero instead of builtin sum() in calculateMinHashScore

### LINKED ISSUE #112: MinHash.calculateMinHashScore uses builtin sum() over a numpy array (~10x on that line)
`MinHash.calculateMinHashScore` sums a numpy boolean array with Python's builtin `sum()`:

**Where:** `mcrit/minhash/MinHash.py:89`

```python
return 100.0 * sum(first_np == second_np) / len(first_np)
```

`sum()` iterates the array element by element in Python. `np.count_nonzero` does the same work in C:

```
sum(a == b)              6.3  us/pair
np.count_nonzero(a == b) 0.65 us/pair
```

~10x on that line, one word changed, no restructuring, no behaviour change (both return the same count; the surrounding float arithmetic is unaffected).

For scale: measured end-to-end scoring throughput in the real code path is **~70,100 pairs/s**, stable to within 2 % across samples spanning a 144x range of pair counts. This line is not the whole cost — tuple unpacking, `np.frombuffer` and call overhead are also in there — so the end-to-end gain will be less than 10x, but the change is free.

**Note:** the full fix for this path is to vectorise the whole batch rather than score pair-by-pair, which is worth ~27–29x rather than a fraction of 10x — see the separate issue on the score histogram for that. This one-word change is worth taking anyway if the larger rewrite is not imminent, and it is a genuinely safe first contribution.

shaurya703:
Fixes #112

`calculateMinHashScore` summed the boolean comparison array with Python's builtin `sum()`, which iterates the numpy array element by element. Switching to `np.count_nonzero` does the same count in C — roughly 10x faster on that line, with identical results.

One-word change, no behaviour difference: both count the number of matching positions, and the surrounding float arithmetic is untouched.

Verified locally that scores are unchanged for the uint8 and uint32 paths (identical inputs → 100.0, one differing word → expected fraction).

### DIFF
--- a/mcrit/minhash/MinHash.py
+++ b/mcrit/minhash/MinHash.py
@@ -86,7 +86,7 @@ class MinHash:
         else:
             first_np = np.frombuffer(first, dtype=np.uint32)
             second_np = np.frombuffer(second, dtype=np.uint32)
-        return 100.0 * sum(first_np == second_np) / len(first_np)
+        return 100.0 * np.count_nonzero(first_np == second_np) / len(first_np)
 
     @staticmethod
     def calculateMinHashIntScore(first, second):


danielplohmann:
Verified on an 11.7 M-function corpus: two full matching runs on this branch (2,940- and 727-function samples, single-process, scalar scoring — i.e. every pair goes through the changed line; 5.0 M scored pairs on the larger sample) produce match reports byte-identical to the recorded reference digests.

One subtlety worth recording: `np.count_nonzero` returns a Python `int` where builtin `sum` returned a numpy scalar, so the score's type shifts from `np.float64` to `float`. The values are bit-identical for every possible match count (both compute the same doubles) and JSON serialisation is unchanged (`np.float64` subclasses `float`) — confirmed by the digest identity above, so this is a non-issue, just documented for the record.

danielplohmann:
thanks for your contribution!

PR #118, opened 2026-08-10
Deterministic tie-break in minhash match reduction (#104)

### LINKED ISSUE #104: Pooled matching is non-deterministic: same sample and config can report different matches
Matching the same sample twice at identical configuration can produce different match reports. This is present on `main` @ `1e7ae59` and reproduces reliably on a large sample.

**Where:** `mcrit/matchers/MatcherInterface.py:171` (`_performMinHashMatching`), reduction at `:202`.

Results are consumed with `pool.imap_unordered` and reduced with:

```python
organized_matching_results[key] = max([original_value, new_value], key=lambda x: x[1])
```

`max` returns the **first** maximal element it sees, so when two candidate functions score identically, which one is reported depends on the order workers happen to finish. `organized_matching_results` is also filled in arrival order, so the ordering inside each function's match list varies too.

**Measured:** four runs of the same sample (win.merlin, 7,018 functions) at identical configuration, `MINHASH_POOL_MATCHING = True`, produced **four different** SHA-256 digests over the `matches` structure, all with identical payload length (26,489,297 bytes):

```
9ff3048b5e4c1fd17f36618bcbcf85cc
5b8945a73378e53e14c8c008a979e96d
6d5ef05cce12f79610455c0e0e664cce
08c9a1f71b9ed1968687a440dd33da89
```

The same sample with `MINHASH_POOL_MATCHING = False` is byte-identical across runs (`45fc8043be1f1bddc1e90a14f9cb0aa6`). Identical length alongside differing content points at list ordering and tie selection rather than at different scores.

**Impact**

1. Re-matching a sample can report a different representative function for a given `(sample_a, function_a, sample_b)` triple. For a tool whose output feeds analyst conclusions, non-reproducibility is a problem in its own right.
2. It makes performance work considerably harder. Any A/B of a change to the matching path has to run single-process to be verifiable at all, because there is no stable baseline to compare a pooled run against. This blocked verifying several optimisations against the shipped configuration, since `MINHASH_POOL_MATCHING` defaults to `True`.

**Fix sketch**

Break the tie deterministically, e.g.

```python
organized_matching_results[key] = max([original_value, new_value], key=lambda x: (x[1], -x[0]))
```

which prefers the lower `function_id_b` on equal score. That happens to be what the single-process path already produces, since it iterates `sorted(candidate_pairs.items())` and then `sorted(candidate_ids)` and `max` keeps the incumbent on a tie — so this change makes the pooled path agree with the non-pooled one rather than inventing a third behaviour.

If the per-function match lists are also expected to be stable, they need an explicit sort at assembly time; arrival order will otherwise still leak through.

**Worth fixing early.** Beyond the correctness argument, a deterministic pooled path is a prerequisite for verifying any future change to matching against the default configuration.

danielplohmann:
Fixes #104.

Both reduction sites in `_performMinHashMatching` now break score ties on the smaller `function_id_b`:

```python
max([original_value, new_value], key=lambda x: (x[1], -x[0]))
```

`max` keeps the first maximal element, so the pooled path previously reported whichever worker's result happened to arrive first through `imap_unordered` — re-matching the same sample could name a different representative function for the same `(sample_a, function_a, sample_b)` triple. The single-process path iterates candidate ids in ascending order (`_unrollGroupsAsPackedTuples` sorts both levels), so "first maximal" already meant "smallest `function_id_b`" there and this change is a behavioural no-op for it.

Verification (11.7 M-function corpus, digests over the full `matches` structure):

- win.merlin (7,018 fns, k=2) — the four-digest repro from #104: two pooled runs now both produce `45fc8043...`, byte-identical to the single-process digest (26,489,297 B payloads).
- win.citadel (2,940 fns, k=1): single-process and two pooled runs all identical (8,077,383 B).
- One end-to-end submit through the real server/queue/worker (pooled, default config) reproduces the single-process digest exactly.
- No assembly-time sort is needed: `_summarizeMatches` pre-populates from `self._function_entries` (deterministic order) and per-function match lists are sorted at assembly (`dict["matches"] = sorted(...)`), so tie selection was the only arrival-order leak.
- 600 tie-stress differential cases (2,109 tie-contested keys) against an independent "smallest function_id_b among maximum scorers" reference: 0 mismatches.

Beyond reproducibility for users, this makes pooled runs verifiable: any future change to the matching path can be A/B-checked against the default `MINHASH_POOL_MATCHING = True` configuration by digest.

### DIFF
--- a/mcrit/matchers/MatcherInterface.py
+++ b/mcrit/matchers/MatcherInterface.py
@@ -199,7 +199,11 @@ class MatcherInterface:
                             key = (sample_id_a, function_id_a, sample_id_b)
                             new_value = (function_id_b, score)
                             original_value = organized_matching_results[key]
-                            organized_matching_results[key] = max([original_value, new_value], key=lambda x: x[1])
+                            # tie-break on the smaller function_id_b so the reported match does
+                            # not depend on worker arrival order (imap_unordered); this matches
+                            # what the single-process path below produces by iterating candidate
+                            # ids in ascending order
+                            organized_matching_results[key] = max([original_value, new_value], key=lambda x: (x[1], -x[0]))
                     if single_batch:
                         self._progress_reporter.step()
         else:
@@ -213,7 +217,7 @@ class MatcherInterface:
                         key = (sample_id_a, function_id_a, sample_id_b)
                         new_value = (function_id_b, score)
                         original_value = organized_matching_results[key]
-                        organized_matching_results[key] = max([original_value, new_value], key=lambda x: x[1])
+                        organized_matching_results[key] = max([original_value, new_value], key=lambda x: (x[1], -x[0]))
                 if single_batch:
                     self._progress_reporter.step()
         full_score_counts = sorted([(item[0] * 64 / 100, item[1]) for item in dict(counted_scores).items()])


danielplohmann:
Additional verification at corpus scale, from an overnight sweep: on **74 samples** (stratified by function count, 246–5,482 functions), one pooled run with this patch produces a digest **byte-identical to the recorded single-process stock digest** for every single sample — 74/74, zero mismatches. The two remaining samples of the 76-sample reference set have no stock reference (they exceeded the reference sweep's 30-minute stock timeout) and were skipped.

So the claim in the PR description holds distribution-wide, not just on the two hand-picked samples: with the tie-break fixed, `MINHASH_POOL_MATCHING = True` and single-process matching are byte-equivalent.

Issue #115, opened 2026-08-10
Smaller storage-layer items: missing indexes, count-before-find, skip() paging, transpiler layering


danielplohmann:
Four smaller storage-layer items found while profiling matching against an 11.7 M-function corpus at `1e7ae59`. Grouping them because none justifies its own thread, and **three of the four are inferred from reading rather than measured** — I have included how to confirm each, so nobody spends effort on a hypothesis of mine without checking it first.

- [ ] **1. Missing indexes on frequently-filtered fields** — `mcrit/storage/MongoDbStorage.py:151` (`_ensureIndexAndUnknownFamily`)

  Present on my instance:
  ```
  functions : _id, function_id, sample_id, _pichash, _picblockhashes.hash, _picblockhashes.offset
  samples   : _id, sample_id
  families  : _id, family_id
  band_N    : _id, band_hash
  ```
  Not indexed but filtered on: `samples.sha256`, `samples.family_id`, `functions.family_id`, `families.family_name`, `functions.function_name`.

  **Inferred, not measured.** To confirm: `db.setProfilingLevel(0, {slowms: 0})`, run one submit, then compare `docsExamined` against `nreturned` per query shape in `system.profile`. A ratio far above 1 confirms a scan. This is the cheapest way to turn the item into a fact, and I never ran it.

- [ ] **2. `getSampleBySha256` counts the whole collection before looking anything up** — `mcrit/storage/MongoDbStorage.py:602`, `count_documents` at `:605`

  A `count_documents(filter={})` non-emptiness test runs before the actual `find_one`, i.e. a full count of `samples` on every submit and every import dedup check. With item 1 fixed the `find_one` becomes an index hit and the count is pure waste. Cheap to fix regardless: `find_one` returning `None` already answers the question the count was asked for.

  **Inferred**; the cost scales with collection size, so it is small at 8,483 samples and would matter on a large instance.

- [ ] **3. `skip()`-based paging is quadratic over a full pass** — `mcrit/storage/MongoDbStorage.py:1149` (`rebuildMinhashBandIndex`) and other `skip()` sites, first at `:587`

  `skip()` is O(n) per page because the server walks and discards the skipped documents, so a full pass over 11.6 M functions is O(n²).

[...]

Issue #109, opened 2026-08-10
_getDb() lazy init is unsynchronised while the server shares one storage across request threads


danielplohmann:
`MongoDbStorage._getDb()` performs an unsynchronised lazy initialisation, and the server shares a single storage object across request threads.

**Where:** `mcrit/storage/MongoDbStorage.py:128` (`_getDb`) and `:141` (`_initDb`).

`_getDb` tests `self._database is None` and assigns only after a full `MongoClient.__init__` has run — a window spanning URI parsing and options construction, i.e. easily wide enough for a thread switch. Two threads arriving cold can therefore both construct a client. The loser's client is dropped without `close()`, so its topology-monitor threads stay alive until garbage collection (pymongo does not close on `__del__`). Both threads also run `_ensureIndexAndUnknownFamily()`, whose

```python
if self.getFamily(0) is None:
    self.addFamily("")
assert self.getFamily(0).family_name == ""
```

is itself non-atomic and ends in an `assert`, so the loser can trip on a duplicated family 0.

**Reachability.** `mcrit/server/application_routes.py` builds **one** `MinHashIndex`, and therefore one `MongoDbStorage`, shared by all request threads; `mcrit/__main__.py` serves with waitress and no `threads=` argument, so the default of 4 applies. Nothing in `Worker.__init__` touches the database, so on a cold server the first concurrent requests race. The matching paths are not affected — they warm `_getDb()` long before anything interesting happens.

**This is inferred from reading, not observed.** I did not manage to trigger it. The window is real but narrow, and the consequences (a leaked client, a duplicate family-0 insert) are individually survivable, which is presumably why it has never been reported.

**Fix sketch**

Double-checked locking:

```python
_db_lock = threading.Lock()

def _getDb(self):
    if self._database is None:
        with self._db_lock:
            if self._database is None:
                self._initDb(...)
    return self._database
```

Two lines, one uncontended lock acquisition per call in the common path. Worth doing on the general principle that a shared mutable singleton behind a multi-threaded server should not be lazily initialised without a lock, rather than because a failure has been seen in the wild.

PR #129, opened 2026-08-14
Make counter init idempotent and enforce unique counters.name (#105)

### LINKED ISSUE #105: counters collection gains a duplicate document on every process start (1,025 observed), and counter selection has no tie-break
The `counters` collection accumulates a duplicate document on every process start. On our instance there are now **1,025 duplicates each** for `query_samples` and `query_functions`.

**Where:** `mcrit/storage/MongoDbStorage.py:166-167`, in `_ensureIndexAndUnknownFamily` (which starts at `:151`).

```python
self._getDb().counters.find_one_and_update(
    filter={"name": "query_samples", "value": 0}, update={"$inc": {"value": 1}}, upsert=True)
```

The intent is "ensure the counter is at least 1, so query ids are never 0". But the filter includes `"value": 0`, so once the counter has advanced past 0 the filter matches nothing — and an `upsert` that matches nothing **inserts a new document** built from the filter's equality terms plus the update. So every process start adds `{name: "query_samples", value: 1}` and `{name: "query_functions", value: 1}`.

**Measured on a live instance** (11.7 M functions, ~1,000 process starts' worth of history):

```
{'families': 1, 'samples': 1, 'functions': 1, 'job': 1,
 'query_samples': 1025, 'query_functions': 1025}      # 2,054 documents in counters
```

The real counters are intact (`query_samples: 242`); the 1,024 extras all sit at `value: 1`.

**Why this is more than cosmetic bloat.** `_useCounterBulk` selects the counter with no sort and no tie-break:

```python
query_result = self._getDb().counters.find_one_and_update(
    filter={"name": name}, update={"$inc": {"value": num_counts}}, upsert=True)
```

With 1,026 documents matching `{"name": "query_samples"}`, which one gets incremented is up to the query planner. It has evidently always picked the original — but nothing in the code or the schema guarantees it, and there is an index on `counters.name`, so the selection is an index scan whose order is not something to rely on across compaction or an index rebuild. If it ever returned one of the `value: 1` duplicates, **query sample and function ids would be handed out again**, colliding with existing `query_samples` / `query_functions` documents.

**Reproduce:** start any process that constructs a `MongoDbStorage` (server, worker, or CLI) against a database whose query counters are non-zero, then `db.counters.find({"name": "query_samples"})`. One new document per start.

**Fix sketch**

Make the initialisation idempotent, e.g.

```python
self._getDb().counters.update_one({"name": "query_samples"}, {"$max": {"value": 1}}, upsert=True)
```

`$max` raises a zero/absent counter to 1 and leaves an advanced one alone, and the filter matches the existing document, so nothing is inserted.

Two follow-ups worth pairing with it:

- a one-off cleanup deleting `{"name": {"$in": ["query_samples", "query_functions"]}, "value": 1}` where a higher-valued document for the same name exists;
- a **unique index on `counters.name`**, which would have made this impossible and makes a regression impossible.

danielplohmann:
Fixes #105.

The counter initialisation in `_ensureIndexAndUnknownFamily` used `find_one_and_update(filter={"name": ..., "value": 0}, ..., upsert=True)`. Once the counter has advanced past 0 the filter matches nothing, and an upsert that matches nothing inserts a new document built from the filter's equality terms plus the update — so every process start added a `{name, value: 1}` duplicate for `query_samples` and `query_functions` (1,025 each measured on our live instance, per the issue). The risk beyond bloat: `_useCounterBulk` selects with a bare `{"name": name}` filter and no tie-break, so with duplicates present the incremented document is whichever the query planner picks — if it ever picked a value-1 duplicate, query sample/function ids would be re-issued and collide with existing `query_samples`/`query_functions` documents. (That it has always picked the original so far is observed behaviour, not guaranteed.)

Three changes, exactly the sketch from the issue:

1. **Idempotent init** — `update_one({"name": ...}, {"$max": {"value": 1}}, upsert=True)`: raises an absent/zero counter to 1, leaves an advanced one alone, and matches the existing document so nothing is ever inserted.
2. **One-off cleanup** (`_removeCounterDuplicates`) — for each query counter name, keeps the highest-valued document (the real counter) and deletes the rest, logging the removal count at INFO. Runs on every start but is a no-op once clean.
3. **Unique index on `counters.name`** — makes a regression structurally impossible. The previous code created this index non-unique, so on existing instances the index is dropped and recreated unique; the cleanup runs first, since the unique build would otherwise fail on polluted databases.

`_useCounterBulk` itself is unchanged: with uniqueness enforced at the schema level there is only ever one document per name to select.

## Verification

- New mongo-backed tests in `tests/testStorage.py` (`MongoDbStorageTest`), covering: (a) constructing storage twice against the same database leaves exactly one counters document per name; (b) a pre-polluted collection (real counter at 242 plus five value-1 duplicates per name, non-unique index) is cleaned to the single highest-valued document and the resulting `name_1` index is unique; (c) counter increments continue from the real value (242 → next id 242, stored value 243).
- Full pytest suite against a throwaway mongod (mcrit-server:1.6.1 image): **93 passed, 5 subtests passed, 4 failed** — the 4 failures are all in `tests/testMatcher.py` and reproduce identically on unmodified v1.6.1 main in the same environment (measured), i.e. pre-existing and unrelated to this change.
- Not verified here: the cleanup against the actual 2,054-document live collection — the numbers above come from the issue; the test reproduces the shape (one real value plus value-1 duplicates) at smaller scale.

---

**Coordination:** touches `_ensureIndexAndUnknownFamily` alongside the storage PRs for #109 and #115 — whichever merges first leaves a trivial textual conflict for the others (different lines of the same method, no semantic interaction).

### DIFF
--- a/mcrit/storage/MongoDbStorage.py
+++ b/mcrit/storage/MongoDbStorage.py
@@ -163,11 +163,19 @@ class MongoDbStorage(StorageInterface):
         self._getDb()["query_functions"].create_index("function_id")
         self._getDb()["query_functions"].create_index("sample_id")
         # ensure that their counters are at least 1, so that they never contain items with sample_id/function_id 0
-        self._getDb().counters.find_one_and_update(filter={"name": "query_samples", "value": 0}, update={"$inc": {"value": 1}}, upsert=True)
-        self._getDb().counters.find_one_and_update(filter={"name": "query_functions", "value": 0}, update={"$inc": {"value": 1}}, upsert=True)
+        # the name-only filter with $max is idempotent: it matches an existing counter instead of upserting a duplicate (#105)
+        self._getDb().counters.update_one({"name": "query_samples"}, {"$max": {"value": 1}}, upsert=True)
+        self._getDb().counters.update_one({"name": "query_functions"}, {"$max": {"value": 1}}, upsert=True)
         self._getDb()["matches"].create_index("match_id")
         self._getDb()["candidates"].create_index("function_id")
-        self._getDb()["counters"].create_index("name")
+        # duplicates must be removed before the unique index can be built on databases polluted by #105
+        self._removeCounterDuplicates()
+        # counters.name must be unique so that _useCounterBulk can never increment a stale duplicate and re-issue ids;
+        # older instances created this index without the constraint, in which case it has to be replaced
+        counter_indexes = self._getDb()["counters"].index_information()
+        if "name_1" in counter_indexes and not counter_indexes["name_1"].get("unique", False):
+            self._getDb()["counters"].drop_index("name_1")
+        self._getDb()["counters"].create_index("name", unique=True)
         self._getDb()["logs"].create_index("username")
         for band_id in range(self._storage_config.STORAGE_NUM_BANDS):
             self._getDb()["band_%d" % band_id].create_index("band_hash")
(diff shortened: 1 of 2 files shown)


danielplohmann:

Issue #69, opened 2024-03-11
Workers consume a lot of ram on query


yankovs:
While doing a query on a sample, often the memory usage of a single worker jumps to 10s of GBs - sometime even more than 60GB. It seems like there's no limit at all and the workers are greedy with memory usage, if some file will require 200GB of ram the worker will try to get it and will probably crash due to lack memory. As a result in a setup with multiple workers it happens quite a lot that some worker will hoard all the memory and starve other workers so that they crash.

Here is a list of hashes for samples that consistently consume a lot of ram on a worker (on our MCRIT instance with 20 million functions):
cea60afdae38004df948f1d2c6cb11d2d0a9ab52950c97142d0a417d5de2ff87
d92f6dd996a2f52e86f86d870ef30d8c80840fe36769cb825f3e30109078e339
bab77145165ebe5ab733487915841c23b29be7efec9a4f407a111c6aa79b00ce
97f1ea0a143f371ecf377912cbe4565d1f5c6d60ed63742ffa0b35b51a83afa2
94433566d1cb5a9962de6279c212c3ab6aa5f18dbff59fe489ec76806b09b15f
a5b38fa9a0031e8913e19ef95ac2bd21cb07052e0ef64abb8f5ef03cf11cb4d5
085b68fa717510f527f74025025b6a91de83c229dc1080c58f0f7b13e8a39904
043aac85af1bda77c259b56cd76e4750c8c10c382d7b6ec29be48ee6e40faa00
84ad84a1f730659ac2e227b71528daec5d59b361ace00554824e0fddb4b453cf
1c4bdd70338655f16cd6cf1eb596cd82a1caaf51722d0015726ec95e719f7a27
29bd1ffe07d8820c4d34a7869dbd96c8a4733c496b225b1caf31be2a7d4ff6df
f72bb91a4569fb9ba2aa40db2499f39bb7aba4d20a5cb5f6dd1e2a9a4ce9af98
9119213b617e203fbc44348eb91150a4db009d78a4123a5cbce6dc6421982a91
a614ed116edc46301a4b3995067d5028af14c8949f406165d702496630cb02ce
0c9edded5ff2ac86b06c1b9929117eab3be54ee45d44fcdb0b416664c7183cbf

I am not sure what is the correct way to handle this, but I think there at least should be a way to limit each worker to some amount of memory.

danielplohmann:
Okay, I have now implemented a first draft for ob processing with subprocesses.
You can try it out by running 
[...]

PR #132, opened 2026-08-14
Pack matching batches to a candidate-pair budget instead of a function count (#107)

### LINKED ISSUE #107: MINHASH_MATCHING_FUNCTION_BATCH_SIZE default of 10000 is an OOM hazard; batch to a pair budget instead
`MINHASH_MATCHING_FUNCTION_BATCH_SIZE` defaults to 10000 (`mcrit/config/MinHashConfig.py:36`). It controls how many query functions are matched per batch, and peak memory scales with it, because scoring materialises every candidate pair as a tuple carrying both 64-byte minhashes.

Measured on win.merlin (7,018 functions, a Go binary) against an 11.7 M-function corpus at `1e7ae59`:

| batch size | max pairs in one batch | peak RSS |
|---|---|---|
| 50 | 1,554,530 | 1,256 MiB |
| 1000 | 12,296,645 | 5,155 MiB |

At the shipped default of 10000 that sample runs as a **single batch** — roughly 61 M pairs at once. Go samples are the pathological case: repeated runtime/stdlib boilerplate creates very large equivalence classes, so candidate pairs per query function are far higher than for typical PE malware (measured means: 8,690 for that Go sample versus 937 and 370 for two non-Go samples of comparable size).

This is a plausible cause of the crashes reported in #69, where a user sees 10s of GB and occasionally >60 GB per worker on a 20 M-function instance. See [this comment](https://github.com/danielplohmann/mcrit/issues/69#issuecomment-5237082862) for the mechanism in detail.

**Fix sketch**

Lowering the default helps immediately, but a fixed function count is the wrong unit in both directions. Candidate volume per query function is extremely skewed — measured on a synthetic 10 M-function corpus, p50 ≈ 7 candidates, p75 ≈ 1,443, p90 ≈ 220,000. So most query functions are nearly free while a small minority carries all the cost, and any constant is simultaneously too large for the tail and needlessly small for the median.

**Adaptive batching** fits the data: `getCandidatesForMinHashes` already returns the candidate volume per query function *before* scoring starts, so batches can be packed to a target *pair* budget rather than a function count. Cheap functions batch in the thousands; the pathological tail goes one at a time. That caps peak memory by construction instead of by guessing a constant.

A caveat worth recording, because I predicted otherwise and was wrong: eliminating the per-pair tuples (by scoring a query function against all its candidates as one numpy block) does **not** make batch size free. With that in place, peak still climbed 1,836 → 6,693 MiB on the same sample from batch 200 → 10000, because the candidate sets and the per-batch signature cache still scale with the batch. It changes the coefficient — 5,155 MiB → 2,982 MiB at batch 1000 — not the shape. So a smaller default (or adaptive batching) is worth doing regardless of any scoring rewrite.

danielplohmann:
Fixes #107.

`MINHASH_MATCHING_FUNCTION_BATCH_SIZE = 10000` bounds the wrong unit: candidate volume per query function spans five orders of magnitude (p50 ≈ 7, p90 ≈ 220k on the measured corpus), so any fixed function count is simultaneously an OOM hazard on the tail and needlessly conservative on the median. At the shipped default, real samples run as a **single batch**: bumblebee (4,557 functions) holds 53.2M pairs at 7.75 GiB peak container RSS on v1.6.1 defaults; #69-class corpora reach tens of GB.

As the issue proposed: `getCandidatesForMinHashes` already knows the exact per-function candidate volume before scoring, so batches are now packed to a **pair budget** (`MINHASH_MATCHING_MAX_PAIRS`). Query functions are banded in small chunks (≤1000, one storage round trip each) and accumulated; whenever pending pairs reach the budget the batch is flushed to cache+scoring. A query function whose candidates alone exceed the budget flushes as a singleton — peak is bounded by the widest single group, which no batching scheme can go below. `0` restores fixed-size batches verbatim. All matcher types inherit the change (the VS intersection and Query cache overrides are untouched hooks).

**Correctness:** results are partition-invariant by #118's determinism. Unit test asserts legacy == small-batches == budget-1 == budget-default on the memory backend; on the real corpus, byte-identical digests against known-goods for four samples across budgets 2M/10M/50M and the eviction-heavy regime.

**Measured trade curve** (bumblebee, 53.2M pairs, union 4.55M functions, v1.6.1 defaults, ABBA where marked):

| arm | wall | peak RSS | mongo reads |
|---|---|---|---|
| legacy single batch (mean of 2) | 248.5 s | 7,751 MiB | 1,138 MiB |
| budget 50M (default) | 252.8 s | 7,320 MiB | 1,258 MiB |
| budget 10M | 417.8 s | 4,654 MiB | 1,545 MiB |
| budget 2M (mean of 2) | 754 s | 3,936 MiB | 2,558 MiB |
| budget 2M, cache sized to union | 370.9 s | 4,799 MiB | 1,256 MiB |

Second sample (merlin, k=2, 53.3M pairs, smaller union): budget 2M = −55% peak (6.67 → 2.98 GiB) at +21% wall — the trade is cheap when the union fits the MatchingCache.

**Why the default is 50M — a guard, not a squeeze.** Tight budgets interact with the persistent MatchingCache: when the budget is far below the candidate union, successive batches evict each other's entries and refetch (the isolation row above shows ~half the 2M slowdown is exactly this). At 50M the budget never binds on normal jobs (+2% wall / −6% peak measured) but caps a 250M-pair runaway at ~5 flushes — roughly 12 GiB scoring-side instead of the >60 GB reported in #69. RAM-constrained deployments can lower it with the curve above as guidance; the config comment documents both the ~250 B/pair rule of thumb and the cache interaction.

Full pytest suite: 96 passed (95 + new partition-invariance test), 5 subtests, 0 failed (mcrit-server:1.6.0 image; the 1.6.1 image carries smda 4.4.5 which breaks 4 testMatcher expectations on unmodified main — separate pre-existing issue).

Evidence tags: all numbers **measured** on the docker-mcrit box, 2026-08-12, raw JSONs preserved (`session-06/results/`).

### DIFF
--- a/mcrit/config/MinHashConfig.py
+++ b/mcrit/config/MinHashConfig.py
@@ -32,8 +32,24 @@ class MinHashConfig(ConfigInterface):
     PICHASH_SIZE: int = 10
     # do not perform minhash matching for pichash matches, instead assume they are implied
     PICHASH_IMPLIES_MINHASH_MATCH: bool = True
-    # size of batches for which candidates are processed
+    # size of batches for which candidates are processed. With MINHASH_MATCHING_MAX_PAIRS
+    # active this only bounds how many query functions are banded per storage round trip;
+    # the memory-relevant unit is the pair budget below.
     MINHASH_MATCHING_FUNCTION_BATCH_SIZE: int = 10000
+    # Upper bound on candidate pairs accumulated before a batch is scored. Candidate volume
+    # per query function is extremely skewed (measured p50 ~7, p90 ~220k candidates), so any
+    # fixed function count is simultaneously too big for the tail (peak RSS scales with
+    # pairs: a single 10000-function batch has been measured holding 50M+ pairs / >7 GiB)
+    # and needlessly small for the median. Packing batches to a pair budget caps matcher
+    # memory by construction; a single query function whose candidate set alone exceeds the
+    # budget is scored in a batch of its own, bounding the peak at the widest single group.
+    # The default is a guard against runaway jobs (#69-class: hundreds of millions of pairs)
+    # and leaves typical jobs in one or two batches (measured +2% wall / -6% peak on a
+    # 53M-pair sample). Lower values buy memory with wall time - measured on that sample:
+    # 10M -> -40% peak / +68% wall, 2M -> -49% peak / +203% wall (roughly ~250 B resident
+    # per in-flight pair; below the candidate-union size, batches also evict each other from
+    # the MatchingCache, which is most of the added wall). 0 restores fixed-size batches.
+    MINHASH_MATCHING_MAX_PAIRS: int = 50000000
     # score each query function against its candidates as one (C, L) numpy block instead of
     # per-pair tuples. Same matches, but pair tuples are never materialised, so peak memory
     # goes from O(pairs) to O(widest candidate group). When enabled, the matching phase runs
--- a/mcrit/config/MinHashConfig.py
+++ b/mcrit/config/MinHashConfig.py
@@ -43,6 +43,9 @@ class MinHashConfig(ConfigInterface):
     # and needlessly small for the median. Packing batches to a pair budget caps matcher
     # memory by construction; a single query function whose candidate set alone exceeds the
     # budget is scored in a batch of its own, bounding the peak at the widest single group.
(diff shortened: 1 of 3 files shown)


danielplohmann:

PR #133, opened 2026-08-14
Survive transient mongod outages in the worker loop; fail jobs of dead children (#106)

### LINKED ISSUE #106: Worker does not survive a mongod restart if a job is in flight
Restarting mongod while a matching job is in flight takes the worker down; it does not reconnect.

```
pymongo.errors.NotPrimaryError: interrupted at shutdown, full error:
{'ok': 0.0, 'errmsg': 'interrupted at shutdown', 'code': 11600, 'codeName': 'InterruptedAtShutdown'}
```

The `spawningworker` child exits non-zero and the job is lost. Any MongoDB maintenance window therefore requires manually restarting the worker.

**It is conditional, which is why it has not been noticed more.** The worker survived **8 consecutive mongod restarts** while idle (container uptime unbroken across all of them) during a benchmarking session. The failure above happened while a job was running. So the diagnosis is "an in-flight operation raises `InterruptedAtShutdown` and nothing catches it", **not** "the connection cannot be re-established" — pymongo reconnects on its own for most errors, and this one escapes to the top level.

**Where:** `mcrit/queue/QueueRemoteCalls.py:351` already catches `Exception` broadly and continues the loop, so the worker *would* survive if the failure were confined to the job. It is not — the child process exits and the parent treats that as fatal.

**Reproduce:** start a matching job on a reasonably large sample, `docker restart <mongo container>` while it runs, observe the worker container exit. Restarting mongod against an *idle* worker does not reproduce it.

**Fix sketch**

Two separable pieces, worth deciding between explicitly:

1. Retry the job's storage calls on `AutoReconnect` / `NotPrimaryError` so a brief mongod outage suspends rather than kills the job.
2. Have the parent treat a non-zero child exit as a *failed job* rather than a fatal condition.

Related: #76 covered the adjacent case where the `singlejobworker` child crashes and the job is nevertheless reported finished with `result_id: None`. That is a different failure (child crash, parent survives, job silently "succeeds"); this one is the parent going down too. Whatever handling came out of #76 is the natural place to hang (2).

danielplohmann:
Fixes #106.

Reproduced the reported behaviour in an isolated stack (scratch mongod + spawningworker + a 4,000-function synthetic report), and it is exactly as conditional as the issue describes:

- **Fast mongod restart (~2 s down), job in flight: worker survives.** pymongo's retryable-write retry absorbs the `InterruptedAtShutdown`; the job loses one attempt and the retry completes. (Matches the issue's "survived 8 consecutive restarts while idle".)
- **Sustained outage (tens of seconds, i.e. a real `docker restart` of a mongod with WT recovery), job in flight: worker dies, exit 1.** Captured traceback: the child dies first; the parent then unwinds and dies at `SpawningWorker.run` → `mongoqueue.next()` → `ServerSelectionTimeoutError` — the poll loop has no exception handling. A second crash follows in `__exit__ → unregisterWorker` during teardown.

The repro surfaced three adjacent latent bugs, all **observed**, all fixed here:

0. **A progress heartbeat racing a release half-locks the job, starving it forever.**
   `Job.progressor()` writes `{"progress": …, "locked_at": now}` unconditionally. If a
   release (orphan reclaim, `release_all_jobs`, `Job.error`) lands between the claim and
   a heartbeat, the job ends with `locked_by: null` but `locked_at` set — and `next()`
   required BOTH null, so the job was permanently invisible to polling. Observed live
   during verification (a job starved exactly this way after an outage; timestamps:
   claim set `started_at`/`locked_at` together, then a heartbeat 1.3 s later rewrote
   `locked_at` alone after an interleaved release). Fixed on both sides: the heartbeat
   only refreshes a lock that is still held (progress is still recorded either way), and
   `next()` treats `locked_by` as the authoritative lock so any pre-existing half-locked
   jobs self-heal — verified live: the starved job was claimed and completed immediately
   after the fix.

1. **A dead child is reported as a finished job** (the #76 ghost, still live in `SpawningWorker`): `_executeJob` logs "Failed Running Remote Job" when the child produced no result_id, but then falls out of the `with job` block cleanly — and `Job.__exit__` with no exception calls `complete()`. The job ends `finished` with `result: None`.
2. **Reader threads race `communicate()` for the child's pipes**: `_executeJobPayload` starts two reader threads on stdout/stderr *and* calls `communicate()`, which reads the same file descriptors. Observed as `OSError: [Errno 9] Bad file descriptor` in `communicate()` when a child died mid-read.

## Changes

- `SpawningWorker.run()` / `Worker.run()` (QueueRemoteCalls): the loop body catches `pymongo.errors.PyMongoError`, logs, and retries with capped exponential backoff (1 s → 30 s, reset on success). A database outage now suspends the worker; pymongo re-establishes the connection when the server returns.
- `queue.clean()` in both `_executeJob` variants was outside the try block; wrapped so periodic housekeeping can't take the worker down (retries next interval).
- `SpawningWorker._executeJob`: a child that exits without a result_id now **raises**, routing the job through `Job.__exit__`'s error path — returned to the queue with `attempts_left` decremented, becoming `failed` after max attempts — instead of being completed empty.
- `SpawningWorker._executeJobPayload`: `communicate(timeout=…)` → `wait(timeout=…)`; the reader threads own the pipes (they already fed the result_id parse). Returns `(result_id, returncode)` so the failure message can carry the child's exit code.
- `SpawningWorker.__exit__`: best-effort teardown — if the DB is unreachable during shutdown, orphaned locks are left for `release_orphaned_jobs`/`clean` to reclaim rather than crashing the exit path.

- `mongoqueue.Job.progressor()`: heartbeat refreshes `locked_at` only while `locked_by` is held; `next()`/`_jobs_to_do()` drop the `locked_at: None` condition (`locked_by` is the lock, `locked_at` its heartbeat) so half-locked jobs self-heal.
- **Own-lock reconciliation**: if the database is down when a job fails, the `Job.__exit__` `error()` write is lost and the job stays locked by a live worker that no longer knows it holds it — unreachable by orphan release (which only covers unregistered consumers). Any job-execution failure now marks the consumer for `release_all_jobs()` before its next claim, replaying the lost release once the database is back. It runs before `next()` so it can never release a freshly claimed job, and is a no-op when the release already landed.

Deliberately NOT done (scope): no per-storage-call retry layer (the issue's option 1). The job in flight during an outage still fails and consumes an attempt; making individual storage calls outage-transparent is a much bigger change, and the queue's existing attempts/retry machinery already covers the gap once the worker survives.

Two more related findings, documented but NOT fixed here (candidates for a follow-up issue):
- `MongoQueue.repair()` computes its staleness threshold as `timedelta(self.timeout)` — that is **300 days**, not 300 seconds, so stale-lock repair effectively never fires.
- A worker that crashes without running `__exit__` leaves its consumer_id registered in `queue_counters.workers`; its locked jobs are then never treated as orphaned (the orphan release only reclaims jobs of UNregistered consumers), so they stay locked until the (broken, see above) `repair()` would act. Worker liveness (e.g. a heartbeat with expiry) would close this properly.

## Verification (all measured, isolated stack, live instance untouched)

- Pre-fix: sustained outage with job in flight kills the worker (traceback above); fast restarts survive. Post-fix, same scenario: worker stays up through a 4+ minute outage, logs capped-backoff retries, resumes polling.
- Post-fix full cycle, zero manual intervention: 150 s outage mid-job → child dies → lock reconciled → job reclaimed, retried, **finished with a real result 24 s after mongod returned** (attempts 3→2, exactly one consumed) → worker alive throughout. A second cycle validated the half-lock self-healing live: a job starved by the progressor race was claimed and completed immediately under the fixed `next()`.
- Unit tests: `tests/testWorkerResilience.py` (run-loop continues across `ServerSelectionTimeoutError`/`NotPrimaryError`; dead child routes to error path; successful child completes) + two mongoqueue regression tests (heartbeat cannot resurrect a released lock; half-released jobs remain claimable). 
- Full suite (mcrit-server:1.6.0 image + throwaway mongod): **100 passed, 5 subtests passed, 0 failed** (95 baseline + 5 new).

Evidence tags: everything in this PR is **measured** except the gunicorn-fork note in the lock comment, which is inferred from deployment layout.

### DIFF
--- a/mcrit/SpawningWorker.py
+++ b/mcrit/SpawningWorker.py
@@ -9,6 +9,8 @@ import time
 import uuid
 from typing import TYPE_CHECKING, Optional
 
+import pymongo.errors
+
 from mcrit.config.McritConfig import McritConfig
 from mcrit.minhash.MinHasher import MinHasher
 from mcrit.queue.QueueFactory import QueueFactory
@@ -56,8 +58,14 @@ class SpawningWorker(Worker):
 
     def __exit__(self, *args):
         # TODO unregister our worker_id from all in-progress jobs found in the queue
-        self.queue.unregisterWorker()
-        self.queue.release_all_jobs()
+        try:
+            self.queue.unregisterWorker()
+            self.queue.release_all_jobs()
+        except pymongo.errors.PyMongoError:
+            # best-effort teardown: if the database is unreachable while we exit, orphaned
+            # locks are reclaimed later by release_orphaned_jobs/clean rather than turning
+            # the shutdown itself into a crash
+            LOGGER.error("Could not release jobs on shutdown, queue cleanup will reclaim them.", exc_info=True)
 
     #### NO REDIRECTION: SPAWM SINGLE JOB WORKERS INSTEAD ###
 
@@ -89,21 +97,10 @@ class SpawningWorker(Worker):
         t2.start()
 
         try:
-            stdout_result, stderr_result = console_handle.communicate(timeout=self._queue_config.QUEUE_SPAWNINGWORKER_CHILDREN_TIMEOUT)
-            stdout_result = stdout_result.strip().decode("utf-8")
-            # TODO: log output from subprocess in the order it arrived
-            # instead of the split to stdout, stderr
-            if stdout_result:
-                LOGGER.info("STDOUT logs from subprocess: %s", stdout_result)
-            if stderr_result:
-                stderr_result = stderr_result.strip().decode("utf-8")
-                LOGGER.info("STDERR logs from subprocess: %s", stderr_result)
-
-            last_line = stdout_result.split("\n")[-1]
-            # successful output should be just the result_id in a single line
-            match = re.match("(?P<result_id>[0-9a-fA-F]{24})", last_line)
-            if match:
-                result_id = match.group("result_id")
+            # the reader threads own the pipes; communicate() would race them for the same
+            # file descriptors (observed as OSError EBADF when a child dies mid-read), so
+            # only wait for the exit code here and let the readers drain the output
+            console_handle.wait(timeout=self._queue_config.QUEUE_SPAWNINGWORKER_CHILDREN_TIMEOUT)
         except subprocess.TimeoutExpired:
             LOGGER.error(f"Job {str(job.job_id)} running as child from SpawningWorker timed out during processing.")
             console_handle.kill()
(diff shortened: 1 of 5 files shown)


danielplohmann:

Issue #134, opened 2026-08-14
Unpinned smda breaks the test suite: 4 testMatcher failures with smda 4.4.5 on unmodified main


danielplohmann:
CI has started failing on every branch and pull request, including ones that touch no matcher code. The cause is outside the repository: `requirements.txt` pins `smda>=4.2.13`, so a fresh install now resolves **smda 4.4.5**, whose disassembly output differs from 4.4.4 in a way that invalidates the hardcoded expectations in `tests/testMatcher.py`.

**Measured** on an unmodified checkout of `main` (`5624676`), same container image and the same throwaway mongod, only the smda version differing:

| smda | result |
|---|---|
| 4.4.4 | `95 passed, 5 subtests passed` |
| 4.4.5 | `4 failed, 91 passed, 5 subtests passed` |

The four are always the same:

```
FAILED tests/testMatcher.py::MatcherTestSuite::testMatcherQuery
FAILED tests/testMatcher.py::MatcherTestSuite::testMatcherSample
FAILED tests/testMatcher.py::MatcherTestSuite::testMatcherVs
FAILED tests/testMatcher.py::MatcherTestSuite::testMatcherVsGroup
```

Each fails as an `AssertionError` comparing the produced match lists against the class-level `function_matches_expected` / aggregation expectations — different `num_bytes`/`num_instructions` and shifted function ids, i.e. the fixtures now disassemble differently, not a matching regression.

**Why this is worth a pin rather than only a fixture refresh:** the expectations encode a specific disassembly of the checked-in `.smda` reports, so any future smda release that changes instruction escaping or function boundaries will break the suite again the moment a fresh environment resolves it — without a single commit to this repository. That makes CI red for contributors who changed nothing related, which is where this was noticed (a set of unrelated storage/queue PRs all went red at once).

**Suggested fix**, in rising order of effort:

1. Pin the tested smda version (`smda==4.4.4`, or `>=4.2.13,<4.4.5`) so CI is reproducible, and bump deliberately.
2. Refresh the four expectations against the newer smda in the same commit that lifts the pin, so the bump is one reviewable change that says "disassembly moved, here is the new baseline".
3. Longer term, consider deriving the matcher expectations from the fixtures at test time (or asserting on relationships rather than exact byte/instruction counts) so the suite tests matching rather than a frozen disassembly.

[...]

PR #135, opened 2026-08-14
Fix repair()'s stale-lock threshold: timedelta days vs seconds

danielplohmann:
`MongoQueue.repair()` computes its staleness threshold as `timedelta(self.timeout)`. `timedelta()`'s first positional argument is **days**, while the timeout is seconds everywhere else (`QUEUE_TIMEOUT`, default 300, passed through `QueueFactory`). The filter therefore asks for locks older than **300 days**, and stale-lock repair could never fire on any realistic instance.

```python
filter={"locked_by": {"$ne": None}, "locked_at": {"$lt": datetime.now() - timedelta(self.timeout)}},
```

Found while reproducing #106 (see #133); reported there as a follow-up and split out here because it stands on its own.

## Scope: this is landmine removal, not a behaviour change

`repair()` has **no callers** in the codebase today (`self.timeout` is used nowhere else), so fixing the unit changes nothing at runtime — which is also what makes it safe to fix on its own.

**Wiring it up is a separate decision, and not a safe one to take casually.** `locked_at` is only refreshed by `Job.progressor()`, and `MatcherInterface` only steps its progress reporter when a job has more than one batch (`if self._num_batches > 1`). At the shipped `MINHASH_MATCHING_FUNCTION_BATCH_SIZE` of 10000, real matching jobs run as a *single* batch — on our corpus one such job takes ~250 s and larger ones considerably more — so they never heartbeat after the initial progress write. A 300 s reclaim threshold would therefore pull the lock out from under a perfectly healthy worker and decrement its `attempts_left`. I have left the method uncalled and recorded that constraint in its docstring rather than guessing at the right design; proper orphan detection probably wants worker liveness (a heartbeat with expiry) rather than a wall-clock lock age.

## Tests

Two cases in `tests/testMongoQueue.py`, both **measured** against a throwaway mongod:

- `test_repair_uses_seconds_not_days`: a lock aged past `timeout` is released (`locked_by`/`locked_at` cleared), `attempts_left` is decremented, and the job becomes claimable again. **This test fails on the unfixed code** (verified by reverting the one-line change: `1 failed, 11 passed`).
- `test_repair_leaves_fresh_locks_alone`: a freshly claimed job keeps its lock and its attempt count.

Full suite: **97 passed, 5 subtests passed, 0 failed** (95 baseline + 2 new), using an environment with smda 4.4.4. The `Unit tests` CI job will show four unrelated `testMatcher` failures here as it does on unmodified `main` — see #134.

---

**Coordination:** touches `mcrit/libs/mongoqueue.py`, as do #127 (`_ensure_indices`, `_jobs_in_progress`) and #133 (`next()`, `Job.progressor()`). All three edit different methods; expect at most a trivial textual conflict for whichever merges later.

### DIFF
--- a/mcrit/libs/mongoqueue.py
+++ b/mcrit/libs/mongoqueue.py
@@ -212,9 +212,18 @@ class MongoQueue:
         """Clear out stale locks.
 
         Increments per job attempt counter.
+
+        NOTE: this currently has no callers. Before wiring it up, be aware that the
+        staleness threshold only means what it says for jobs that refresh their lock:
+        `locked_at` is only bumped by Job.progressor(), and a matcher running as a single
+        batch never steps its progress reporter, so a long job can look "stale" while it
+        is healthily running. Reclaiming it would decrement attempts_left underneath a
+        live worker.
         """
         self._getCollection().find_one_and_update(
-            filter={"locked_by": {"$ne": None}, "locked_at": {"$lt": datetime.now() - timedelta(self.timeout)}},
+            # timedelta()'s first positional argument is DAYS: the timeout, which is
+            # seconds everywhere else (QUEUE_TIMEOUT, default 300), has to be named
+            filter={"locked_by": {"$ne": None}, "locked_at": {"$lt": datetime.now() - timedelta(seconds=self.timeout)}},
             update={"$set": {"locked_by": None, "locked_at": None}, "$inc": {"attempts_left": -1}},
         )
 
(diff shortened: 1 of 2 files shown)


danielplohmann:
CI note: the `Unit tests` failures here are the same four `tests/testMatcher.py` cases that fail on unmodified main because smda is unpinned (#134) — my two new queue tests pass, and the deselected/passed counts match the baseline. Ruff and Build package are green.

Issue #137, opened 2026-08-17
Split `_xcfg` out of the functions collection (4.58x less I/O on the matching path)


danielplohmann:
Proposal: move the `_xcfg` blob out of `functions` into its own collection keyed by `function_id`, so the frequently-read function metadata (minhash, pichash, sample/family ids, offsets) no longer shares documents with the rarely-read disassembly.

**Note up front:** this is not a request for someone to pick up. The work is already laid on our side — migration script, a full rehearsal against production data, and an implementation plan with the touch points enumerated — and we intend to implement it ourselves. Filing it as an issue so the design and the numbers are on the record before a schema-changing PR appears, and so nobody duplicates the effort.

## Why

After the #113/#114-class optimisations landed, the dominant remaining cost of matching is fetching candidate minhashes: a **186 B projection out of documents averaging 4,188 B**, where `_xcfg` is most of the weight. WiredTiger pages whole documents, so the working set for matching is effectively the entire collection (48.9 GB logical, 18.4 GB on disk) rather than the ~1.3 GB actually read.

Profiling at level 2 shows the fetch is already a perfectly shaped IXSCAN with `docsExamined == nreturned` — there is no waste in planning, and **no index can improve it**. Only lighter documents can.

Matching queries themselves do not change: `_getCacheDataForFunctionIds` already projects `_xcfg` away. The entire gain is page density. A useful consequence for review: **match reports must come out byte-identical**, so any digest difference in a PR implementing this is a bug rather than an expected change.

## Measured, on an 11.7 M-function corpus

Full rehearsal of the migration against production data (read-only source, migrated shape written to a scratch database on the same mongod):

| | before | after |
|---|---|---|
| `functions` storage | 18,353 MB | **3,130 MB** (5.9x smaller) |
| average document | 4,188 B | **721 B** |
| new `xcfg` collection | — | 14,210 MB (+130 MB `_id` index) |
| total storage | 18,353 MB | 17,340 MB (slightly **less**) |
| bytes read into WT cache per matching job | 37,778 MiB | **8,248 MiB** (4.58x less) |
| known-good match digests | `f3c4aeba…` | **identical** |

Total storage does not grow — homogeneous documents compress better than mixed ones — so the split needs transient headroom during migration, not permanent disk.

[...]

PR #141, opened 2026-08-20
Split `_xcfg` out of the functions collection (#137)

danielplohmann:
Implements #137.

`_xcfg` moves out of `functions` into its own collection keyed by function id (`_id` **is** the function id, so no secondary index is needed and the lookup is the primary key). Query functions mirror this in `query_xcfg`, keeping the existing `functions` / `query_functions` separation rather than mixing sign conventions in one collection.

**Matching queries do not change.** `_getCacheDataForFunctionIds` already projected the blob away, so the entire gain is page density — which is also why match reports must come out byte-identical, and any digest difference here would be a bug rather than an expected change.

## Measured on an 11.7 M-function corpus

| | before | after |
|---|---|---|
| `functions` storage | 18,353 MB | **3,130 MB** (5.9x smaller) |
| average document | 4,188 B | **721 B** |
| bytes read into WT cache per matching job | 37,778 MiB | **8,248 MiB** (4.58x less) |
| total storage | 18,353 MB | 17,340 MB (slightly **less**) |
| known-good match digests | `f3c4aeba…` | **identical** |

Whole-job wall-clock gain remains the earlier controlled **1.10–1.48x** (Amdahl: scoring, pichash filtering and result assembly are untouched). The 4.58x is measured as WiredTiger `bytes read into cache` after a restart, which is cache-state-independent; a first wall-clock reading of ~2.8x was discarded as contaminated by OS page-cache asymmetry (the migrated copy fits in RAM, the original does not).

## Verification

- **124 tests pass** (117 + 7 new in `MongoDbXcfgSplitTest`), under smda 4.4.4, 4.4.5 and 4.4.7.
- **Matching digests** against a fully migrated 11.7 M-function database: citadel `f3c4aeba…` and photoloader `30a3e58a…`, both match.
- **Blob-reader equivalence** — the check that actually exercises this change. Old code against the unsplit database vs new code against the migrated one, comparing hashes of what callers receive:

| sample | functions | `getFunctionsBySampleId` | `getFunctionById(with_xcfg)` | unique blocks |
|---|---|---|---|---|
| citadel 601 | 850 | identical | identical | identical |
| photoloader 7455 | 40 | identical | identical | identical (empty result) |
| bumblebee 8263 | 4,557 | identical | identical | identical |

- **Migration and rollback** rehearsed on a 40,000-function copy of live data: migrate 2.8 s, rollback 3.3 s, and a fingerprint over every `(function_id -> blob)` pair is identical before and after a full cycle. The copy path was separately rehearsed at full scale: 40.3 GB in 20.1 minutes.

## Where review should look hardest

Writing this, the existing test suite caught three things my plan had wrong. They are the parts most worth a second pair of eyes:

1. **The reader audit.** My plan listed four readers and missed three. `getUnhashedFunctions` feeds *minhash computation* — the indexing path needs the disassembly — and `getFunctionsBySampleId` feeds both `Worker.calculateMinHashes` and the code-reference links in match reports, while `getFunctions` had no projection at all before and so carried the blob too. All now go through one `_attachXcfgBlobs` helper. **If any consumer of `FunctionEntry.xcfg` is still unaccounted for, that is the bug class to look for.**
2. **The `None` vs `{}` contract.** `testFunctionHandling` asserts that after `deleteXcfgForSampleId`, `with_xcfg=True` yields `{}` rather than `None`. That distinction existed *because* the old code blanked the field in place instead of removing it, and it is preserved by decoding a missing blob as `{}`.
3. **`_id` uniqueness.** `functions.function_id` is a non-unique index, so duplicate ids were previously tolerated; an `_id`-keyed collection makes them a `DuplicateKeyError`. `clearStorage` therefore has to drop the new collections in lockstep with resetting the counters — and any other code path that resets counters needs the same treatment.

## Not established, stated plainly

- **`inplace` at full scale.** Rehearsed at 40,000 functions, not 11.7 M. The copy path was measured at full scale and `inplace` does strictly less work per document, but the number is not measured.
- **A restored-dump rehearsal.** Rollback works, but the recommended production sequence is still dump → migrate → verify, because rollback depends on the blob collection being intact.
- **Concurrency.** Nothing here tests migrating while a worker indexes new samples; the migration should run with writers stopped.

## Operator-facing notes for the release

- The migration is required: this is a schema change, and a database that has not been migrated will read empty disassembly.
- `STORAGE_DROP_DISASSEMBLY` now genuinely reclaims space. `deleteXcfgData` drops the collection instead of blanking every document, which is what that flag always implied.
- Suggested release-note line: take a dump before running `--mode inplace`, and run it with the worker stopped.

### DIFF
--- a/mcrit/index/MinHashIndex.py
+++ b/mcrit/index/MinHashIndex.py
@@ -424,7 +424,8 @@ class MinHashIndex(QueueRemoteCaller(Worker)):
         return infos
 
     def getStatus(self, with_pichash=True):
-        storage_stats = self.getStorage().getStats(with_pichash=with_pichash)
+        storage = self.getStorage()
+        storage_stats = storage.getStats(with_pichash=with_pichash)
         status = {
             "status": {
                 "db_state": storage_stats["db_state"],
@@ -437,6 +438,15 @@ class MinHashIndex(QueueRemoteCaller(Worker)):
                 "num_pichashes": storage_stats["num_pichashes"],
             }
         }
+        # operator signal for a half-migrated instance (#137): the inline fallback keeps serving
+        # correct, but the split's space reclamation only happens once migrate_xcfg_split ran.
+        # Omitted rather than False when a backend cannot determine it cheaply.
+        inline_xcfg_remaining = None
+        storage_inline_check = getattr(storage, "hasInlineXcfgRemaining", None)
+        if callable(storage_inline_check):
+            inline_xcfg_remaining = storage_inline_check()
+        if inline_xcfg_remaining is not None:
+            status["status"]["inline_xcfg_remaining"] = inline_xcfg_remaining
         return status
 
     def getVersion(self):
--- a/mcrit/index/MinHashIndex.py
+++ b/mcrit/index/MinHashIndex.py
@@ -440,11 +440,9 @@ class MinHashIndex(QueueRemoteCaller(Worker)):
         }
         # operator signal for a half-migrated instance (#137): the inline fallback keeps serving
         # correct, but the split's space reclamation only happens once migrate_xcfg_split ran.
-        # Omitted rather than False when a backend cannot determine it cheaply.
-        inline_xcfg_remaining = None
-        storage_inline_check = getattr(storage, "hasInlineXcfgRemaining", None)
-        if callable(storage_inline_check):
-            inline_xcfg_remaining = storage_inline_check()
+        # Every storage answers this - the interface default is None ("not applicable") - so the
+        # field is simply omitted when a backend cannot determine it.
+        inline_xcfg_remaining = storage.hasInlineXcfgRemaining()
         if inline_xcfg_remaining is not None:
             status["status"]["inline_xcfg_remaining"] = inline_xcfg_remaining
         return status
--- a/mcrit/migrations/migrate_xcfg_split.py
+++ b/mcrit/migrations/migrate_xcfg_split.py
@@ -27,7 +27,7 @@ import sys
 import time
 from datetime import UTC, datetime
 
-from pymongo import ASCENDING, MongoClient, UpdateOne
+from pymongo import ASCENDING, MongoClient, ReplaceOne, UpdateOne
 
 from mcrit.config.McritConfig import McritConfig
 
@@ -52,10 +52,29 @@ def put_state(target_db, state):
     target_db[STATE_COLLECTION].replace_one({"_id": state["_id"]}, state, upsert=True)
 
 
+def recreate_indexes(source_db, target_db, collection_name):
+    """Rebuild every secondary index, carrying over its options.
+
+    Only key order and `unique` would survive a hand-rolled copy; TTL, sparse,
+    partialFilterExpression and friends live in the spec alongside them and are lost if
+    not forwarded - so forward everything the server reports except the fields
+    `create_index` derives itself.
+    """
+    for index in source_db[collection_name].list_indexes():
+        keys = list(index["key"].items())
+        if keys == [("_id", 1)]:
+            continue
+        options = {key: value for key, value in index.items() if key not in ("key", "v", "name", "ns")}
+        target_db[collection_name].create_index(keys, **options)
+        log("  recreated index %s %s" % (keys, options or ""))
+
+
 def copy_verbatim(source_db, target_db, batch_size):
     names = [n for n in source_db.list_collection_names() if n in VERBATIM_COLLECTIONS or n.startswith(VERBATIM_PREFIXES)]
     for name in sorted(names):
-        if target_db[name].estimated_document_count() >= source_db[name].estimated_document_count():
+        # exact counts, not estimated_document_count(): the estimate reads cached collection
+        # metadata and can misread a partially copied target as "already present"
+        if target_db[name].count_documents({}) >= source_db[name].count_documents({}):
             log("verbatim %s already present, skipping" % name)
             continue
         target_db[name].drop()
@@ -70,12 +89,7 @@ def copy_verbatim(source_db, target_db, batch_size):
         if batch:
             target_db[name].insert_many(batch, ordered=False)
             copied += len(batch)
-        # recreate the indexes the matching path relies on
-        for index in source_db[name].list_indexes():
-            keys = list(index["key"].items())
-            if keys == [("_id", 1)]:
-                continue
-            target_db[name].create_index(keys, unique=index.get("unique", False))
+        recreate_indexes(source_db, target_db, name)
         log("verbatim %s: %d documents" % (name, copied))
 
 
@@ -115,7 +129,10 @@ def migrate_functions(source_db, target_db, mode, batch_size, limit, source_coll
             # copy the function documents themselves, minus the blob
             function_documents = list(source_db[source_collection].find({"function_id": {"$in": function_ids_done}}, {"_xcfg": 0}))
             if function_documents:
-                target_db[source_collection].insert_many(function_documents, ordered=False)
+                # upserts, not insert_many: a resume can re-enter the batch whose inserts
+                # landed before the state save was killed - plain inserts would die on
+                # duplicate _id there and make copy mode unresumable
+                target_db[source_collection].bulk_write([ReplaceOne({"_id": document["_id"]}, document, upsert=True) for document in function_documents], ordered=False)
         elif mode == "inplace":
             # only now that the xcfg documents are durable, drop the blobs. Restricted to
             # ids we just wrote, so a crash can never unset an xcfg that was not saved.
@@ -139,12 +156,7 @@ def migrate_functions(source_db, target_db, mode, batch_size, limit, source_coll
         # without these the target is not a like-for-like copy and any benchmark against it
         # measures missing indexes rather than the schema change
         index_started = time.perf_counter()
-        for index in source_db[source_collection].list_indexes():
-            keys = list(index["key"].items())
-            if keys == [("_id", 1)]:
-                continue
-            target_db[source_collection].create_index(keys, unique=index.get("unique", False))
-            log("  recreated index %s" % keys)
+        recreate_indexes(source_db, target_db, source_collection)
         log("indexes rebuilt in %.1f s" % (time.perf_counter() - index_started))
 
     elapsed = time.perf_counter() - started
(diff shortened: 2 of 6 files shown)


danielplohmann:
Agreeing with all five, and the resumability one is the one I'm glad was caught — the docstring
promised something `insert_many` cannot deliver, and `ReplaceOne(..., upsert=True)` is the right fix
because it makes the function-document path match the idempotency the xcfg path already has. #4 is
fair and slightly embarrassing: `verify()` was deliberately switched to `count_documents({})` after
`estimatedDocumentCount()` proved 36,662 docs stale on this instance, and then the estimate was left
in `copy_verbatim`'s skip gate.

One thing not in the list, which I'd rate above #1 because it affects serving rather than the
migration tool.

**Both blob readers discard an inline `_xcfg` instead of falling back to it.**

```python
function_document["_xcfg"] = blobs.get(function_document["function_id"], "{}")            # _attachXcfgBlobs
function_document["_xcfg"] = self._fetchXcfgBlobs([function_id]).get(function_id) or "{}" # getFunctionById
```

If a function still carries its disassembly inline and has no blob yet, the reader throws the inline
value away and returns `{}`. That is not reachable once a migration has completed, but it is very
reachable before one, and there is a circularity: `migrate_xcfg_split.py` ships inside the package,
so the new code has to be deployed in order to run it. Between deploying and finishing — or if
someone upgrades the image and restarts without migrating at all — every un-migrated function reads
as empty disassembly:

- `Worker.calculateMinHashes`, reached via `getUnhashedFunctions`, computes minhashes from nothing
  and writes those signatures into the index
- match reports lose their code references
- `getUniqueBlocks` returns nothing

All silently, because `{}` is a legitimate value that means "no disassembly stored".

One line per reader removes the whole class:

```python
function_document["_xcfg"] = blobs.get(fid) or function_document.get("_xcfg") or "{}"
```

That makes readers correct before, during and after the migration, and tolerant of a rollback to the
pre-`$unset` state, which also removes the undocumented "migrate before serving traffic" requirement
the PR currently implies. Worth pairing with a cheap startup check —
`functions.find_one({"_xcfg": {"$exists": True}})` — logging loudly when inline blobs remain, so a
half-migrated instance announces itself instead of quietly returning empty results.

This also sharpens observation #5: leftover placeholders are harmless because their inline *value* is
already `"{}"`, not because overwriting inline data is safe. The general case is not.

Separately, I went looking for leaked query blobs and there aren't any — `deleteSample` collects the
query function ids and calls `_deleteXcfgForFunctionIds` before `query_functions.delete_many`, and
that is the only deletion path for them. That part is clean.

danielplohmann:
The inline fallback and the four reader paths look right, and #143's `{architecture: fingerprint}`
change is better than what I asked for — keeping `/status` a scalar while only the persisted export
format becomes a mapping means mcritweb is unaffected, and the non-dict branch even covers a stray
string from a branch build.

But the startup probe needs rethinking, and since I asked for it without specifying a cost bound,
that is on me.

`_warnOnInlineXcfgRemaining()` runs `functions.find_one({"_xcfg": {"$exists": True}})`. There is no
index on `_xcfg`, so this is a COLLSCAN that **cannot terminate early precisely when everything is
healthy** — a fully migrated instance has no matching document, so the scan runs to completion.
Measured on this corpus (11,628,392 functions, pre-migration 18,353 MB collection):

| state | probe cost |
|---|---|
| inline blobs still present (migration incomplete) | **0.003 s** — early exit on the first hit |
| fully migrated (nothing matches) | **191.3 s** — full scan |

Post-split the collection is 5.9x smaller (3,130 MB per the rehearsal), so call it very roughly
**~30 s** once migrated. The incentives are exactly inverted: the broken state is free to detect, the
correct state pays.

The per-process guard does not help, because the guard is module-level state and the jobs are not
in-process. `SpawningWorker` runs each job as
`subprocess.Popen(["python", "-m", "mcrit", "singlejobworker", ...])`, and `SingleJobWorker` builds
its own storage via `StorageFactory.getStorage(config)` → `MongoDbStorage.__init__` →
`_ensureIndexAndUnknownFamily()` → the probe. **Fresh OS process per job means the scan happens per
job.** For scale: in the 1.5.0-vs-1.6.2 sweep, 1.6.2 jobs at stock k=2 ran 0.8–7 s each, so a ~30 s
preamble would dominate a typical job by an order of magnitude.

The second cost is worse than the latency. Scanning the whole functions collection walks ~3 GB
through the WiredTiger cache on every job, evicting the minhash and band working set that matching
depends on. That is a direct attack on what #137 exists to achieve — the rehearsal's 4.58x reduction
in bytes read per job assumes the hot collection stays resident.

Two suggestions, either of which is enough:

1. **Take it out of `_ensureIndexAndUnknownFamily()`.** That method runs on *every* storage
   construction, including read-only tooling — it is already the reason indexes once got created
   unintentionally on a production instance. A migration-completeness warning is an operator
   concern: surface it in `/status`, or in an explicit CLI check, not in a constructor on the job
   path.
2. **If it should stay automatic, make the common case O(1).** `xcfg.estimated_document_count()`
   versus `functions.estimated_document_count()` are both metadata reads; only run the real probe
   when those are far apart. `migrate_xcfg_split` also already records progress in
   `c1_migration_state`, which is a tiny collection and authoritative about whether the migration
   finished. Either way, bound the fallback with `max_time_ms` and log "could not determine" on
   timeout rather than blocking startup.

Worth noting the estimate caveat cuts both ways here: `estimatedDocumentCount()` was 0.3% stale on
this instance, which is fine for a trigger that decides whether to look harder, and not fine as the
answer itself.

danielplohmann:
Moving it off the job path is the right fix, and recording *why* in the docstring is worth more than
the fix itself — that reasoning is exactly what would otherwise get re-litigated in six months.

One consequence of the 5 s bound though, which inverts the same way the original probe did.

`getStatus()` now calls `hasInlineXcfgRemaining()` on every request, and in the **migrated** state the
probe cannot finish inside the budget. Measured on this instance: the full COLLSCAN takes **191.3 s**
on the pre-split collection (18,353 MB, 11,628,392 functions). Post-split the collection is 3,130 MB
per the rehearsal, so scaling gives roughly **33 s** — still ~6.5x over the 5 s cap.

So once an instance is fully migrated:

- the probe always times out, so `hasInlineXcfgRemaining()` always returns `None`
- `None` means the field is omitted, so `/status` never reports migration state at all
- and each `/status` call still burns 5 s of collection scan and the attendant WiredTiger cache churn

which is better than per-job, but it means the check cannot answer the question it exists for
precisely when the answer is "yes, migrated" — and mcritweb polls `/status`.

The asymmetry that makes this fixable: the probe is only ever *needed* in the un-migrated case, where
it exits at the first hit in **0.003 s**. So the migrated case should be decided without scanning at
all:

- `migrate_xcfg_split` already maintains `c1_migration_state` (`put_state`, keyed per phase). It
  records a resume cursor but no terminal marker. Writing one when a phase drains, and reading it
  first, is O(1) and definitive.
- Failing that, `xcfg.estimated_document_count()` vs `functions.estimated_document_count()` are both
  metadata reads; only fall through to the probe when they are far apart. The estimate's ~0.3%
  staleness on this instance is fine for a trigger that decides whether to look harder, though not
  as the answer itself.

With either, the bounded probe becomes a rarely-hit safety net instead of the normal path, and
`/status` can report a definitive `false` rather than omitting the field — worth doing, since
"migration complete" is a thing an operator actively wants to see, not merely the absence of a
warning.

danielplohmann:
Separate from the timing point, `1f2c40a` has a bug that this PR's own CI will not catch.

`StorageInterface.hasInlineXcfgRemaining()` is declared as `raise NotImplementedError`, and
`MemoryStorage` does not override it. `getStatus()` guards with:

```python
storage_inline_check = getattr(storage, "hasInlineXcfgRemaining", None)
if callable(storage_inline_check):
    inline_xcfg_remaining = storage_inline_check()
```

An inherited abstract method satisfies both `getattr` and `callable`, so the guard does not fire -
it only protects a storage class that does not inherit from `StorageInterface`. The call then raises
`NotImplementedError` out of `getStatus()`, so **`/status` breaks on the memory backend**.

It stays invisible on this branch because nothing here calls `getStatus()` under memory storage -
`testMinHashIndex.py` only covers construction and the candidate-pair generator. But #143 adds
`testStatusReportsEscaperProvenance`, which calls `index.getStatus(with_pichash=False)` on the test
config (memory-backed - it passes in the `not mongo` job without a server). So this goes red as soon
as both PRs are in, and whichever merges second looks like the culprit.

Cleanest fix is to make the interface default return `None` rather than raise. The docstring already
defines `None` as "not applicable or could not be determined cheaply", which is exactly what a
backend without the pre-split shape should say - and then the `getattr`/`callable` dance in
`getStatus()` can go away too, since every storage answers the question.

One correction on my previous comment while I am here: the ~33 s post-split figure is scaled from the
measured 191.3 s by collection size (18,353 MB -> 3,130 MB), not measured directly. A COLLSCAN's cost
is partly per-document rather than purely per-byte, so if the split collection is fully resident the
real number could be closer to 6-10 s. The argument does not depend on which: `/status` is polled, so
a full-collection scan per call churns the WiredTiger cache that matching depends on whether it takes
5 s or 33 s, and the O(1) trigger removes it either way. I can measure it properly once I migrate the
benchmark copy - which also closes the "inplace at full scale unrehearsed" gap flagged in the
original C1 decision doc.

danielplohmann:
The terminal-marker approach in `12ae4fb` is the right shape — O(1) for the healthy case, scoped to
`inplace` only (a rehearsal `copy` says nothing about this instance), no marker on `--limit` runs, and
`unsplit`'s `delete_many` regex removes it on rollback. The two new tests pin the part that matters:
the marker wins over a probe that would answer differently.

Just noting that the `NotImplementedError` issue is **still live at `12ae4fb`** — that commit touches
`migrate_xcfg_split.py`, `MongoDbStorage.py` and `testStorage.py`, so the interface default is
unchanged and `MemoryStorage` still does not override it. Reproduced on this exact head:

```
$ python -c "
from tests.context import config
from mcrit.index.MinHashIndex import MinHashIndex
MinHashIndex(config).getStatus(with_pichash=False)"
...
NotImplementedError
```

`storage: MemoryStorage` → `getStatus` raises. One line in `StorageInterface` (`return None` instead
of `raise NotImplementedError`) closes it, and then the `getattr`/`callable` guard in `getStatus()`
can go too.

danielplohmann:
`829e863` closes the last of it — interface default answers `None`, `getStatus()` calls it directly
now that the `getattr`/`callable` guard is unnecessary, and the new `MemoryStorageTest` case asserts
both halves that matter: no exception, and the field *omitted* rather than reported as `False`. My
reproduction from the previous comment passes on this head.

Independently verified the branch at `829e863` on a full copy of a production corpus (8,689 samples /
11,665,054 functions), rather than only reading the diffs:

- **suite**: 67 passed in the `not mongo and not sleep` job, **132 passed** in the full job with a
  mongo service (main is 117, so the PR's additions all run)
- **digest**: `bench/run_match.py 601 --k 1 --batch 200` under the deployment config produced
  `f3c4aebadf8b9be7af56306b71f2c7a8` — **byte-identical to the recorded known-good** for that
  sample/config

The digest run is the one worth noting: the copy is **un-migrated**, so it exercised the new inline
fallback on every reader path rather than the post-split path. That is the code that had no coverage
before this round, and it reproduces the known-good exactly — 4,631,845 pairs scored, same result.

No further findings from me on this PR.

Issue #142, opened 2026-08-21
Minhashes silently go stale when smda's escaper changes: no provenance, no selective repair, no test


danielplohmann:
## Summary

Minhashes are derived from smda's escaped instruction representation, but nothing in MCRIT records
or checks which smda version produced a stored minhash. When smda's escaper changes, an index
silently accumulates stale signatures: they still look valid, they still match each other, and
newly submitted samples containing the *same code* no longer find them.

Measured on a production instance (8,689 samples / 11,628,392 functions, indexed under smda ≤ 4.3,
now running 4.4.7). Unbiased random sample of 15,974 functions, recomputing each stored function's
signature from its own stored `_xcfg`:

| | |
|---|---|
| stored signatures that differ from what the running smda computes | **19.50%** (~2.27M) |
| mean self-similarity for **identical** code | 97.61% |
| below `MINHASH_MATCHING_THRESHOLD` of 50 | 0.22% |
| **lose all 20 bands → unretrievable at any threshold or `BAND_MATCHES_REQUIRED`** | **0.19%** (~21,800) |

That last row is the real cost: those functions are invisible to candidate retrieval, so no
threshold or `k` tuning recovers them. Verified end-to-end against the live `band_*` collections —
they return in 20/20 bands under their stored signature and 0/20 under the signature today's smda
computes from the very same stored disassembly.

Cause is upstream and isolated to smda 4.4.5 (filed as danielplohmann/smda#277): a correct fix
stopped escaping segment-qualified memory operands (`gs:[0x60]`, `fs:[0x30]`, `es:[edi]`) as
`CONST`. 99.6% of the drift traces to it. This will happen again — escaping should keep improving.
What is missing here is our side: detection and targeted repair.

## Gap 1: the index cannot tell it is stale

- `MinHashConfig.getConfigHash()` covers `MINHASH_STRATEGY`, `FN_MIN_INS`, `FN_MIN_BLOCKS`,
  `SIGNATURE_LENGTH`, `SIGNATURE_BITS`, `MINHASH_SEED`, `PICHASH_SIZE`;
  `ShinglerConfig.getConfigHash()` covers weight strategy, logbuckets, weights, seed. **Neither
  includes the smda version**, even though the escaper is as much an input to a minhash as the
  seed is.
- `SampleEntry` *does* store a per-sample `smda_version`, so the information needed to detect a
  heterogeneous index is already in the database — it is simply never compared against anything.

[...]

PR #143, opened 2026-08-21
Record which escaper produced a minhash (#142)

danielplohmann:
Addresses the detection half of #142. Upstream counterpart: danielplohmann/smda#277.

## Why

MinHashes are derived from smda's escaped instruction representation — the shinglers build their
tokens from `getMnemonicGroup()` and `getEscapedOperands()`. That makes the escaper as much an
input to a minhash as `MINHASH_SEED`, but it lives in a separately versioned package and **no
config hash covers it**. When smda changes how it escapes, stored minhashes silently stop being
comparable to newly computed ones: they still look valid, still match each other, and identical
code submitted afterwards no longer finds them.

smda 4.4.5 is the worked example — it stopped escaping segment-qualified memory operands
(`gs:[0x60]`, `fs:[0x30]`, `es:[edi]`) as `CONST`. Measured on a corpus of 11,628,392 functions
indexed under smda ≤ 4.3 and now running 4.4.7:

| | |
|---|---|
| stored signatures that differ from what the running smda computes | **19.50%** |
| **lose all 20 LSH bands → unretrievable at any threshold or `k`** | **0.19%** (~21,800) |

Verified by querying the band index directly: those functions return in **20/20** bands under the
signature they were indexed with and **0/20** under today's. The smda change is correct — nothing
announced it.

## What this adds

**`mcrit/minhash/EscaperFingerprint.py`** runs the escaper over a fixed, committed probe of
instructions covering the operand and mnemonic classes escaping distinguishes, and hashes the
result. It needs no report, binary or disassembly — only `SmdaInstruction` built from four-field
lists — so it is computable anywhere, cheaply (~470 µs).

Validated against real releases, which is the point of the exercise:

| smda | fingerprint | `gs:[0x60]` escapes to |
|---|---|---|
| 4.2.13, 4.3.10, 4.4.2, 4.4.4 | `32243755bdcdbc75` | `M REG, CONST` |
| 4.4.5, 4.4.6, 4.4.7 | `110c9d152232a2e6` | `M REG, PTR` |

It detects exactly the boundary that moved, and nothing else in that range.

**`getStatus()`** now reports `smda_version` and `escaper_fingerprint`, so an operator can see
which escaper an instance indexes with.

**`getExportData()`** records both alongside the shingler and minhash hashes that already gate
reusability ("config hashes need to be identical in order for minhashes to be reusable" — the
escaper belongs in that set). **`addImportData()`** warns and sets `escaper_mismatch` when an
export was escaped differently. It *warns rather than refuses*, deliberately: a given escaper
change affects only a share of functions, so the data stays useful, unlike a config-hash mismatch.
Exports predating the field are unaffected, and the fingerprint degrades to `"unavailable"` rather
than raising if smda exposes no escaper we can drive (below the `smda>=4.2.13` floor).

No behaviour change to matching, hashing or storage — this is provenance only, so it is
results-neutral and fits a patch release. No version bump, matching how #141 was prepared.

## Tests

13 new (130 passed total, from 117 on main), covering determinism, hex shape, probe coverage,
graceful degradation when the escaper cannot be resolved, that the fingerprint **reacts** to an
escaper change, that the probe keeps covering the segment-qualified-memory and far-pointer operand
classes (otherwise it would not have caught 4.4.5), and the export/import provenance paths
including a legacy export with no escaper field.

`ruff format --check .` and `ruff check .` clean.

## Not in this PR

Selective repair — picking affected samples by `smda_version` and rehashing only those, instead of
`recalculateMinHashes()` wiping every minhash and dropping all 20 band collections up front (which
leaves an instance unable to match until the rebuild completes). That touches the same storage
paths as #137/#141, so it is better done after that lands. Tracked in #142.

### DIFF
--- a/mcrit/index/MinHashIndex.py
+++ b/mcrit/index/MinHashIndex.py
@@ -17,7 +17,7 @@ from mcrit.index.SearchCursor import FullSearchCursor, MinimalSearchCursor
 from mcrit.index.SearchQueryParser import SearchQueryParser
 from mcrit.libs.utility import compress_encode, decompress_decode
 from mcrit.matchers.MatcherQueryFunction import MatcherQueryFunction
-from mcrit.minhash.EscaperFingerprint import FINGERPRINT_UNAVAILABLE, getEscaperFingerprint
+from mcrit.minhash.EscaperFingerprint import FINGERPRINT_UNAVAILABLE, getEscaperFingerprint, getEscaperFingerprints
 from mcrit.minhash.MinHash import MinHash
 from mcrit.minhash.MinHasher import MinHasher
 from mcrit.queue.QueueFactory import QueueFactory
@@ -152,8 +152,10 @@ class MinHashIndex(QueueRemoteCaller(Worker)):
                 "shingler": self._shingler_config.getConfigHash(),
                 "minhash": self._minhash_config.getConfigHash(),
                 # minhashes derive from smda's escaped instructions, so the escaper is an input
-                # just like the seeds above - see mcrit/minhash/EscaperFingerprint.py
-                "escaper": getEscaperFingerprint(),
+                # just like the seeds above - see mcrit/minhash/EscaperFingerprint.py. Persisted
+                # per architecture, so widening the probe later cannot invalidate exports that
+                # are already in the wild.
+                "escaper": getEscaperFingerprints(),
                 "smda_version": SMDA_VERSION,
             },
             # family mapping is needed for function_entries as they only contain family_ids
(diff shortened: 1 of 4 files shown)


danielplohmann:
All four accepted. Small nuance on #3: the `except Exception` path does emit `LOGGER.warning(...)`, so
it is not entirely silent — but a warning raised inside a `/status` call is easy to miss, so the
suggestion to assert per-instruction probe output in the tests stands.

#1 has something underneath it worth settling now rather than later, because this value gets
**persisted**.

`getExportData()` stores the fingerprint as a single opaque string produced from the default
architecture tuple, currently `("intel",)`, and `addImportData()` compares that string. So the moment
a later version widens the default — adding an aarch64 or cil probe, which is exactly what #1 asks
for — the combined hash changes for everyone, and **every export produced before that point mismatches
on import**. The warning would fire saying the imported minhashes are not comparable, when intel
escaping had not changed at all. False positives are the worst failure mode for a diagnostic: they
teach operators to ignore it, which costs more than the missing coverage does.

Cheapest fix is to store a mapping rather than a scalar, while nothing in the wild depends on the
current shape:

```python
"escaper": {"intel": "110c9d152232a2e6"},   # architecture -> fingerprint
```

and compare only the intersection of the keys both sides have. Adding an architecture later is then
non-breaking, older exports keep comparing correctly on the architectures they do carry, and the
mismatch message gets more precise as a side effect — "escaping changed for intel, unchanged for
aarch64" localises the problem in a way a single hash never can.

That also makes #1 itself cheap to land incrementally: an arm probe can be added whenever someone
with the domain knowledge writes realistic operands for it, without invalidating anything already
exported.

I'd rather fold this into the same amend than ship a persisted format I already know we will want to
change.

Issue #151, opened 2026-08-30
Family statistics drift from the collections they summarise, and /status reports the drifted values as the corpus size


danielplohmann:
`getStats` derives `num_samples` and `num_functions` by summing the denormalised per-family counters rather than counting the collections (`mcrit/storage/MongoDbStorage.py:1503-1506` on `main`):

```python
for family_document in self._getDb().families.find():
    stats["num_families"] += 1
    stats["num_samples"] += family_document["num_samples"]
    stats["num_functions"] += family_document["num_functions"]
```

The comment above it - "use family statistics to derive relevant values" - makes clear this is deliberate, and it is the right call for cost: counting 11.7M documents on every `/status` poll is not viable. The problem is what is on the other end. Those counters are maintained by non-atomic increments spread over several write paths, some of which can lose an update with no error, and **nothing anywhere recomputes them**. `/status` is their only visible consumer, so drift stays invisible until someone compares against the collections.

### Measured

Live instance on 1.8.1. Actual corpus by `countDocuments({})`: **8,691 samples / 11,669,208 functions**. `/status` reported **8,485 / 11,701,622** - 206 samples short and 32,414 functions over.

Per family:

* **125 families under-count** `num_samples`, **1 over-counts**; net **205** samples short.
* 45 families disagree with the summed `statistics.num_functions` of their own member samples.
* One `family_id` (1908) is referenced by a sample but has **no family document at all**.

### The write paths

Five things, roughly in order of how easily each loses an update.

1. **`_updateFamilyStats` has no `upsert`** (`mcrit/storage/MongoDbStorage.py:626-636`):

   ```python
   def _updateFamilyStats(self, family_id, num_samples_inc, num_functions_inc, num_library_samples_inc):
       self._getDb().families.update_one(
           {"family_id": family_id},
           {"$inc": {...}},
       )
   ```

   If no family document matches, the increment is silently discarded - no exception, no log line. This is the defect that turns any of the others into permanent drift.

2. **`importSampleEntry` never ensures the family exists** (`:844-852`). It increments using `sample_entry.family_id` taken straight from the imported entry, whereas `addSmdaReport` calls `addFamily` first (`:830`) before incrementing (`:838`). Importing a sample whose family document is not present in the target instance therefore lands in case 1 and loses the increment.

3. **`modifySample`'s family relabel writes absolute values computed from an earlier read** (`:661-680`) instead of using `$inc`:

   ```python
[...]

Issue #149, opened 2026-08-30
deleteSample leaves empty band documents behind, and its upsert can create them


danielplohmann:
`deleteSample` removes a sample's function ids from the band index with `$pull`, which empties the `function_ids` array but never removes the document. It also passes `upsert=True` on that path, so pulling a band hash that is *not* present creates a document instead of doing nothing.

Neither is a correctness problem for matching - an empty posting list contributes no candidates - but both leave permanent residue in the band collections, and the second means a delete can *add* documents to the index.

This is distinct from #25, which was the original bug of leaving dangling `function_id`s behind. That was fixed; this is the residue of the fix.

### The code

`MongoDbStorage._updateBands` (`mcrit/storage/MongoDbStorage.py:1077` on `main`) issues the same `UpdateOne` for both directions:

```python
if method == "push":
    update_command = {"$push": {"function_ids": {"$each": function_ids}}}
else:
    update_command = {"$pull": {"function_ids": {"$in": function_ids}}}
band_updates.append(UpdateOne({"band_hash": band_hash}, update_command, upsert=True))
```

`deleteSample` reaches it at `mcrit/storage/MongoDbStorage.py:609` via `self._updateBands(minhashes_to_remove, method="pull")`.

### Measured

On a throwaway database, mirroring exactly what the delete path issues:

```python
db.band_x.insert_one({"band_hash": 111, "function_ids": [1, 2, 3]})
db.band_x.bulk_write([
    UpdateOne({"band_hash": 111}, {"$pull": {"function_ids": {"$in": [1, 2, 3]}}}, upsert=True),
    UpdateOne({"band_hash": 999}, {"$pull": {"function_ids": {"$in": [7]}}}, upsert=True),
])
```

leaves two documents:

```
{'band_hash': 111, 'function_ids': []}     <- emptied, not removed
{'band_hash': 999}                         <- created by the upsert, no function_ids at all
```

On a real instance (11.6M functions, 20 bands): deleting three samples - 2,621 of whose functions carried a minhash, occupying **25,368** distinct `(band, band_hash)` pairs - left **3,052 tombstones**. Determined by rebuilding the band index afterwards and counting how many of those band hashes had vanished entirely, i.e. how many of them the deleted functions had been the sole member of.

### Inferred, not measured

[...]

PR #154, opened 2026-08-30
Answer getUniqueBlocks from an inverted picblockhash index (92s -> 0.11s)

danielplohmann:
`getUniqueBlocks` decides "is this block hash unique to the requested samples?" by reading **every function in the database that carries picblockhashes**. On this Malpedia-sized instance (8,691 samples / 11,669,208 functions) that is **9,088,495 documents and 51,364,030 block entries** streamed into Python, **~92 s per call** — and the cost does not depend on how many samples were asked about, because it scans the corpus either way.

This adds a `picblockhashes` collection and answers the same question by reading one document per *candidate* hash.

```js
{_id: "<block hash>", sample_ids: [<sample_id>, ...]}
```

### Measured, on the live corpus, identical results both ways

| request | candidates | scan (today) | index | |
|---|---|---|---|---|
| 1 sample (`win.citadel`) | 4,095 | 92–97 s | **0.11 s** | 1 unique block |
| 5 samples | 155,849 | 107 s | **2.12 s** | 59,977 unique blocks |

12,139,968 distinct hashes, **0.47 GB + 0.25 GB for the `_id` index**, 245 s to build.

## The decisions, since they are the reviewable part

**Why this shape.** The sample list is the whole question. Function ids and offsets would multiply the document count by the ~5.6 blocks each function averages, and `getUniqueBlocks` already holds them from the candidate side. Family ids would make a relabel invalidate the index for no gain.

**Why not the existing `_picblockhashes.hash` index instead.** It answers the same query and needs no new collection — I measured it as the cheaper option first. But its `$in` fan-out pulls whole 708-byte function documents, so it degrades as the request grows: **12.7x at one sample, 2.3x at five**, against 407x and 37x here, on the same runs. The sample-set case is the one that matters — [fkie-cad/mcritweb#140](https://github.com/fkie-cad/mcritweb/pull/140) is about to give Unique Blocks its own page taking a set of samples, where today the only caller passes a list of one.

**Why maintaining this on the write path is safe, when #149 and #151 are both bugs in exactly that.** The difference is idempotency:
- add uses `$addToSet` with `upsert`, which has **set semantics** — a retried or replayed insert converges instead of drifting. #151's family counters use `$inc`, where a lost or repeated update is permanent and needs a recount to find.
- delete scopes the `$pull` by `_id`, taking the hashes from the sample's own functions in one indexed read. So **no multikey index on `sample_ids` is needed** — that would have cost 0.24 GB over 12.1M documents. The consequence is that the hook must run *before* the function documents are deleted, which is why it sits where it does in `deleteSample`.
- emptied posting lists are deleted rather than left. They would actually read correctly (an empty list means "held by nobody", so the candidate survives), but #149 is that same residue in the band index and there is no reason to repeat it.
- a family reassignment is a no-op, pinned by a test so that a future change adding `family_id` here has to confront it.
- `recalculateAllPicHashes` rewrites block hashes wholesale, so it marks the index stale rather than trying to diff old against new per function — which is what the rebuild does anyway, in one pass.

**Why a completeness flag rather than "the collection exists".** A half-filled index reports blocks as unique that are not — a wrong answer, not a slow one. So the index is read only when a settings flag vouches for it. A database with no functions gets the flag for free (an empty index over an empty corpus is complete), so **fresh instances maintain it from the first submit and never need a rebuild**. An existing database upgrading into this does *not* get the flag and keeps using the scan — which is retained, not deleted — until an operator calls the rebuild. **Upgrading changes performance, never results.** While the flag is unset the write hooks skip their work, since there is no point maintaining an index nothing reads.

**`clearStorage` drops the collection.** Leaving it would keep asserting that hashes are held by samples that no longer exist.

## Exposure

`Worker.rebuildPicBlockHashIndex` (`@Remote(progress=True)`, so `MinHashIndex` forwards it automatically), a `GET /rebuild_picblockhash_index` route, and `McritClient.rebuildPicBlockHashIndex` — mirroring `rebuild_index` / `recalculate_pichashes`. `MemoryStorage` gets a `return 0` implementation with a comment: it holds every function in a dict already, so its elimination has nothing to index, and the backends stay interchangeable.

## Tests

Five new mongo-backed tests, all using two samples built from the same report so they carry identical block hashes:

- **index and scan agree** — the same `getUniqueBlocks` call answered both ways, for one sample and for two
- maintained on add and delete, including that a delete leaves no emptied posting lists
- the rebuild reproduces exactly what the write hooks built
- family reassignment leaves the index untouched
- nothing is written while the index is incomplete, and the rebuild recovers it

```
208 passed, 34 subtests passed in 85.72s
All checks passed!            # ruff check
136 files already formatted   # ruff format --check
```

## What I did not do, and what I could not verify here

- **No version bump or changelog entry**, to avoid conflicting with the other open PRs — fold into whichever release suits.
- **The `$out` rebuild replaces the collection atomically** but holds no lock against concurrent writes; a submit landing mid-rebuild could be missed. It sets the flag afterwards, so the window is "rebuild then submit" and the fix would be to re-run it. Worth a maintainer's opinion on whether that needs more than a docstring.
- **Maintenance costs were measured per-operation on a scratch copy** (add 0.52 s/sample, delete 0.40 s/sample), not under concurrent load through the real write paths.
- The scale figures come from this one corpus on one machine.

### DIFF
--- a/mcrit/Worker.py
+++ b/mcrit/Worker.py
@@ -264,6 +264,11 @@ class Worker(QueueRemoteCallee):
 
     # Reports PROGRESS
     @Remote(progress=True)
+    def rebuildPicBlockHashIndex(self, progress_reporter=NoProgressReporter()):
+        return self._storage.rebuildPicBlockHashIndex(progress_reporter=progress_reporter)
+
+    # Reports PROGRESS
+    @Remote(progress=True)
     def recalculatePicHashes(self, progress_reporter=NoProgressReporter()):
         return self._storage.recalculateAllPicHashes(progress_reporter=progress_reporter)
 
--- a/mcrit/client/McritClient.py
+++ b/mcrit/client/McritClient.py
@@ -105,6 +105,15 @@ class McritClient:
             return response
         return handle_response(response)
 
+    def rebuildPicBlockHashIndex(self):
+        """
+        Schedule a job that rebuilds the inverted picblockhash index getUniqueBlocks reads; answers the job id
+        """
+        response = requests.get(f"{self.mcrit_server}/rebuild_picblockhash_index", headers=self.headers)
+        if self.raw:
+            return response
+        return handle_response(response)
+
     def repairMinHashes(self):
         """
         Schedule a job that rehashes only the samples whose minhashes an older smda escaper produced (#142); answers the job id
--- a/mcrit/server/StatusResource.py
+++ b/mcrit/server/StatusResource.py
@@ -95,6 +95,13 @@ class StatusResource:
         return
 
     @timing
+    def on_get_rebuild_picblockhash_index(self, req, resp):
+        """Schedule a job that rebuilds the inverted picblockhash index getUniqueBlocks reads. Answers the job id."""
+        index_report = self.index.rebuildPicBlockHashIndex(force_recalculation=True)
+        resp.data = jsonify({"status": "successful", "data": index_report})
+        db_log_msg(self.index, req, "StatusResource.on_get_rebuild_picblockhash_index - success.")
+        return
+
     def on_post_recompute_family_stats(self, req, resp):
         """Schedule a job that sets every family's sample, function and library counters from the collections, recreating missing family documents. Answers the job id."""
         job_id = self.index.recomputeFamilyStats(force_recalculation=True)
--- a/mcrit/server/application_routes.py
+++ b/mcrit/server/application_routes.py
@@ -106,6 +106,10 @@ def get_app():
     # schedule a job that sets every family's sample/function counters from the collections (#151)
     _app.add_route("/recompute_family_stats", status_resource, suffix="recompute_family_stats")  # post
     # schedule a job that updates all function_entries where the pichash was possibly calculated with an outdated SMDA version
+    # schedule a job that rebuilds the inverted picblockhash index getUniqueBlocks reads.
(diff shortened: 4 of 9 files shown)


danielplohmann:
Rebased onto `main` after #171/#172/#173/#174/#180 landed. All three gates clean on the rebase: **260 passed**, ruff check + format, `ty` 0.0.74.

**One conflict needed a judgement rather than a merge.** My side of `deleteSample` was the pre-#174 body, which `main` has since rewritten twice — #174 made it decrement family stats by what `delete_many`/`delete_one` actually removed, and #180 factored the band pull into `_pullBandEntries`. Taking my side back would have silently reverted both. So I kept `main`'s `deleteSample` and inserted only the one line this PR needs:

```python
function_minhashes = self._getFunctionMinHashesBySampleId(sample_id)
self._pullBandEntries(function_minhashes)
# drop this sample from the picblockhash index while its functions still exist to be read
self._removeSampleFromPicBlockHashIndex(sample_id)
```

The ordering constraint from the original PR still holds and is why the hook sits there: it reads the sample's own functions, so it has to run before they are deleted.

The other three conflicts were adjacent-edit collisions where both sides were needed — `rebuildPicBlockHashIndex` next to #180's `repairMinHashes` and #174's `recomputeFamilyStats` in `McritClient`/`StatusResource`, and both fake-collection helpers in `testMongoDbStorageDeleteSample`.

PR #163, opened 2026-09-06
Let a function be renamed, and record who named it (fkie-cad/mcritweb#72)

### LINKED ISSUE #72: allow modifying samples, families, function
_(the original issue had no description)_

---

### Migration note (2026-08-04)

Re-checked against current `master`: samples and families **can** be modified from the UI by now (`explore.modifySample` / `explore.modifyFamily`, edit modals in `templates/table/sample_row.html` and `templates/table/family_row.html`, backed by the `modifySample`/`modifyFamily` backend jobs).
Functions **cannot** - there is no `modifyFunction` path anywhere in the code base. Scope of this issue is therefore reduced to functions.

---

<sub>Migrated from the project's previous, private repository `danielplohmann/mcritweb` (issue #136), opened 2022-10-13 by [blattm](https://github.com/blattm).</sub>

r0ny123:
Closes fkie-cad/mcritweb#72 (modify functions), together with its MCRITweb half, fkie-cad/mcritweb#173.

## What changed

Families and samples could be modified through the API, functions could not: no route, no client call, no storage operation.

- `PUT /functions/{id}` with `function_name` sets the name on both storage backends and records it as a `FunctionLabelEntry` by the requesting user (the `username` header), unless that user already submitted that label — the same shape `updateFunctionLabels` writes for labels arriving with a submitted report, so a function's label history is one list whichever way a name got there. An empty name clears the name and records nothing. Query functions (negative ids) answer 404, as do unknown ids; a non-printable or over-long name answers 400.
- `MinHashIndex.modifyFunction` is synchronous: unlike `modifySample`/`modifyFamily` it touches one document and no statistics, so there is nothing a job would keep consistent.
- `McritClient.modifyFunction(function_id, function_name)`.

## Tests

Storage tests on both backends (name set, label recorded once per user, anonymous without a user, empty name clears, unknown/query ids refused, other functions untouched), resource tests for the validation and the pass-through of the user.

## Verified against a running instance

`modifyFunction(273, "decrypt_config_v2")` as `alice` → `GET /functions/273` shows the name and a label `('decrypt_config_v2', 'alice')`; a non-ASCII name answers 400 and an unknown id 404. The MCRITweb side renames from the function page and shows the label history.

## In MCRITweb (fkie-cad/mcritweb#173)

the rename modal

### DIFF
--- a/mcrit/client/McritClient.py
+++ b/mcrit/client/McritClient.py
@@ -348,6 +348,16 @@ class McritClient:
         if data is not None:
             return FunctionEntry.fromDict(data)
 
+    def modifyFunction(self, function_id: int, function_name: str):
+        """
+        Set the name of the function <function_id>; the name is also recorded as a label by this client's username.
+        Supported by mcritweb API pass-through
+        """
+        response = requests.put(f"{self.mcrit_server}/functions/{function_id}", {"function_name": function_name}, headers=self.headers)
+        if self.raw:
+            return response
+        return handle_response(response)
+
     ###########################################
     ### Matching
     ###########################################
--- a/mcrit/index/MinHashIndex.py
+++ b/mcrit/index/MinHashIndex.py
@@ -3,7 +3,7 @@ import json
 import logging
 import re
 from datetime import datetime, timedelta
-from typing import Dict, List
+from typing import Dict, List, Optional
 
 from smda.common.SmdaReport import SmdaReport
 from smda.SmdaConfig import SmdaConfig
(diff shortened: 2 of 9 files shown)


danielplohmann (approved):
Approved. Synchronous by design and the comment says why; the 'not yet labelled by this user' condition sits in the update's filter, so two concurrent identical requests cannot both push the label. Composed with #168 and #169 and run on a deployment corpus: 323 passed, 49 subtests, ruff + format + ty clean.

PR #164, opened 2026-09-06
Keep the match flags booleans so they survive the round trip (#155)

### LINKED ISSUE #155: MatchedFunctionEntry.getMatchTuple() corrupts the match flags, on a served route
Found this while auditing MCRITweb against mcrit v1.8.1. It's small but it's on a route you actually serve, so any API consumer sees wrong data.

## What happens

`MatchedFunctionEntry.__init__` stores the *masked bits* under boolean-sounding names:

```python
self.match_is_minhash = match_tuple[4] & MatcherFlags.IS_MINHASH_FLAG   # 0 or 1
self.match_is_pichash = match_tuple[4] & MatcherFlags.IS_PICHASH_FLAG   # 0 or 2
self.match_is_library = match_tuple[4] & MatcherFlags.IS_LIBRARY_FLAG   # 0 or 4
```

and then `getMatchTuple()` multiplies each of them by its flag *again*:

```python
self.match_is_minhash * MatcherFlags.IS_MINHASH_FLAG
  + self.match_is_pichash * MatcherFlags.IS_PICHASH_FLAG
  + self.match_is_library * MatcherFlags.IS_LIBRARY_FLAG
```

So the pichash bit comes back as 2×2 = 4, and the library bit as 4×4 = 16.

## Round trip, executed against v1.8.1

```
flags in -> flags out
   0 -> 0
   1 -> 1
   2 -> 4    <-- pichash-only match now reads as IS_LIBRARY
   3 -> 5
   4 -> 16   <-- library match now has no flag set at all
   5 -> 17
   6 -> 20
   7 -> 21
```

Six of the eight possible values are wrong. The two that survive do so by accident (0 and 1, because 1×1 = 1).

The two that matter in practice:

- **flags 2** (pichash-only) becomes **4**, which is `IS_LIBRARY_FLAG` — a plain pichash match is reported as a library match.
- **flags 4** (library) becomes **16**, which is no flag at all — the library bit is silently lost.

## Why it's not just latent

`MinHashIndex.getMatchesFunctionVs` builds the entry and serialises it directly:

```python
"match_entry": MatchedFunctionEntry(function_id_a, function_entry_a.binweight,
                                    function_entry_a.offset, match_tuple).toDict(),
```

`mcrit/index/MinHashIndex.py:441`

and that's served on `GET /matches/function/{function_id}/{function_id_b}` (`mcrit/server/application_routes.py:141` → `MatchResource`). That line has been there a long while, so this isn't new — it just doesn't show up in the normal matching path, where the flags never make the round trip.

MCRITweb proxies the same route verbatim, so anything reading it through the API gets the corrupted flags. MCRITweb's own function-vs page only renders `matched_score`, which is why nobody noticed from the UI.

## Fix

Either store real booleans in `__init__` (`bool(match_tuple[4] & FLAG)`), or drop the second multiplication in `getMatchTuple()` and just OR the stored masks together. The second is the smaller change and can't be got wrong by a future caller reading `match_is_library` expecting a bool.

Worth a round-trip test over all eight values — that's what makes it obvious.

r0ny123:
Closes #155.

`MatchedFunctionEntry` stored `flags & IS_*_FLAG` — a masked int — under `match_is_minhash/pichash/library`, and `getMatchTuple()` multiplied each by its flag value again. A pichash-only match (2) came back as 4, a library match, and a library match (4) as 16; only 0 and 1 survived. That tuple is what `GET /matches/function/{a}/{b}` serves.

The attributes are now the booleans their names and annotations promise. The test round-trips all eight flag combinations through `toDict()`/`fromDict()` and checks the three attributes are real booleans; it fails on `main` for six of the eight.

### DIFF
--- a/mcrit/storage/MatchedFunctionEntry.py
+++ b/mcrit/storage/MatchedFunctionEntry.py
@@ -35,9 +35,11 @@ class MatchedFunctionEntry:
         self.matched_sample_id = match_tuple[1]
         self.matched_function_id = match_tuple[2]
         self.matched_score = match_tuple[3]
-        self.match_is_minhash = match_tuple[4] & MatcherFlags.IS_MINHASH_FLAG
-        self.match_is_pichash = match_tuple[4] & MatcherFlags.IS_PICHASH_FLAG
-        self.match_is_library = match_tuple[4] & MatcherFlags.IS_LIBRARY_FLAG
+        # stored as booleans: getMatchTuple multiplies them with the flag values again, and a
+        # masked int there would square the flag and corrupt the round trip (#155)
+        self.match_is_minhash = bool(match_tuple[4] & MatcherFlags.IS_MINHASH_FLAG)
+        self.match_is_pichash = bool(match_tuple[4] & MatcherFlags.IS_PICHASH_FLAG)
+        self.match_is_library = bool(match_tuple[4] & MatcherFlags.IS_LIBRARY_FLAG)
         self.matched_family = None
         self.matched_link_score = 0
         self.matched_unique = None
--- a/tests/testDataObjects.py
+++ b/tests/testDataObjects.py
@@ -8,8 +8,10 @@ from copy import deepcopy
 
 from smda.common.SmdaReport import SmdaReport
 
+import mcrit.matchers.MatcherFlags as MatcherFlags
 from mcrit.minhash.MinHash import MinHash
 from mcrit.storage.FunctionEntry import FunctionEntry
+from mcrit.storage.MatchedFunctionEntry import MatchedFunctionEntry
 from mcrit.storage.MatchingResult import MatchingResult
 from mcrit.storage.SampleEntry import SampleEntry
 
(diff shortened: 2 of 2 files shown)


danielplohmann:
Closing in favour of #182, which fixes this by construction rather than by coercion.

Both PRs fix #155 in `MatchedFunctionEntry`, and they conflict — I tested it: merging either makes the other conflict in `MatchedFunctionEntry.py`. So one has to go, and #182 is the better base:

* it keeps the flags as the integer the wire format carries and exposes `match_is_minhash/pichash/library` as read-only views, so `getMatchTuple()` cannot corrupt them — there is no masked-int-multiplied-by-flag step left to get wrong, rather than a `bool()` coercion in front of it;
* it also fixes a second round-trip bug found on the way (`toDict` emitted `matches.functions` as a dict where `fromDict` reads a list, so `fromDict(toDict())` raised `TypeError`);
* it carries the `__slots__` change with it: 217 → 145 B/pair retained, measured on a 240k-pair report.

Nothing here is wrong — this is a de-duplication, not a rejection. Thanks for the fix; it survives inside #182.

PR #168, opened 2026-09-06
Select jobs before paging them, and count what a listing would show (fkie-cad/mcritweb#57, fkie-cad/mcritweb#51)

### LINKED ISSUE #51: Proper sort / search for jobs
- [ ] sort should be done in mcrit and not only consider displayed elements
- [ ] search bar wouldn't hurt
- [ ] ... what about cross jobs? Implement and mcritweb search/sort engine?

---

### Migration note (2026-08-04)

Re-checked against current `master`: sorting and pagination are done in the backend by now, via `client.getQueueData(start=..., limit=..., method=..., state=..., ascending=...)` in `data.jobs()`.
The search bar in `templates/jobs.html:143-146` POSTs a `Search` field which `data.jobs()` reads into `query` and forwards to the template, but it is never used for filtering - so job search is currently a no-op that only echoes `Results for "<term>"`. Cross jobs still need consideration.

---

<sub>Migrated from the project's previous, private repository `danielplohmann/mcritweb` (issue #102), opened 2022-08-11 by [blattm](https://github.com/blattm).</sub>

### LINKED ISSUE #57: META: Jobs
#### Cross Job
- [x] `danielplohmann/mcritweb#105` (closed)
- [ ] #46 
- [x] `danielplohmann/mcritweb#80` (closed)
- [x] `danielplohmann/mcritweb#63` (closed)
- [x] `danielplohmann/mcritweb#62` (closed) 
#### Others
- [x] `danielplohmann/mcritweb#39` (closed)
- [ ] #51
- [x] #47

### Frontend
- [ ] #55 
- [ ] #41
- [ ] #39
- [x] `danielplohmann/mcritweb#28` (jobs page was since reworked into per-method categories, so there is no "Others" tab anymore)

---

<sub>Migrated from the project's previous, private repository `danielplohmann/mcritweb` (issue #112), opened 2022-08-17 by [blattm](https://github.com/blattm).</sub>

r0ny123:
**Stacked on #161** (base branch `fix/37-record-job-owner`); only the last two commits are this PR.

Closes fkie-cad/mcritweb#51 and closes fkie-cad/mcritweb#57 (the Jobs tracker) on the backend side: `getQueueData` applied its text filter to a page *after* the page had been cut, so `start=0, limit=25, filter=x` answered "the matches among jobs 0–24" instead of "the first 25 matches", and nothing could say how many jobs matched at all. A state was decided in Python per document, i.e. the whole queue was loaded for a filtered page.

## What changed

- `MongoQueue._job_query` builds one query from method, state, text filter and requesting user, and `get_jobs` pages *that*. Each state is a query of its own, in the precedence order of `_identifyJobState`; a test inserts every combination of the fields that classification reads (32 documents) and asserts each state query selects exactly the documents the Python classification does.
- The text filter is a literal, case-insensitive match on the method name and the serialized parameters (what `Job.parameters` renders from), regex characters escaped.
- `get_job_count` answers the same selection, exposed as `GET /jobs/count` and `McritClient.getQueueCount()`; `GET /jobs` accepts `filter` and `username`, and `getQueueData` gains `username`. The client URL-encodes the free-text filter (it did not before).
- `LocalQueue.get_jobs` used to receive `ascending` in its `filter` parameter, by position; it has the same selection now.

## Verified against a running instance

`getQueueCount()`, `getQueueCount(username="alice")`, `getQueueCount(filter="SampleVs")`, `getQueueCount(state="finished")` and a filtered, user-narrowed page all answer the expected subsets; `GET /jobs/count?username=alice` returns `{"count": 1}`.

### DIFF
--- a/mcrit/client/McritClient.py
+++ b/mcrit/client/McritClient.py
@@ -580,43 +580,49 @@ class McritClient:
             return response
         return handle_response(response)
 
-    def getQueueData(self, start=0, limit=0, method=None, filter=None, state=None, ascending=False):
+    @staticmethod
+    def _job_selection_query(method=None, filter=None, state=None, username=None, **more):
+        """The query string of the parameters that select jobs, URL-encoded (a filter is free text)."""
+        params = {"method": method, "filter": filter, "state": state, "username": username, **more}
+        present = {key: value for key, value in params.items() if value is not None and value is not False and value != 0}
+        return "?" + urllib.parse.urlencode(present) if present else ""
+
+    def getQueueData(self, start=0, limit=0, method=None, filter=None, state=None, ascending=False, username=None):
         """
-        Get queue data, optionally from <start> and <limit> many
+        Get queue data, optionally from <start> and <limit> many, narrowed to a <method>, a
+        <state>, jobs whose parameters contain <filter> (case-insensitive) and/or jobs requested
+        by <username>. The narrowing is applied before paging, so a page is a page of the matches.
         Supported by mcritweb API pass-through
         """
-        query_string = "?ascending=True" if ascending else ""
-        if isinstance(start, int) and start > 0:
-            if len(query_string) == 0:
-                query_string = f"?start={start}"
-            else:
-                query_string += f"&start={start}"
-        if isinstance(limit, int) and limit > 0:
-            if len(query_string) == 0:
-                query_string = f"?limit={limit}"
-            else:
-                query_string += f"&limit={limit}"
-        if isinstance(method, str) and method is not None:
-            if len(query_string) == 0:
-                query_string = f"?method={method}"
-            else:
-                query_string += f"&method={method}"
-        if isinstance(filter, str) and filter is not None:
-            if len(query_string) == 0:
-                query_string = f"?filter={filter}"
-            else:
-                query_string += f"&filter={filter}"
-        if isinstance(state, str) and state is not None:
-            if len(query_string) == 0:
-                query_string = f"?state={state}"
-            else:
-                query_string += f"&state={state}"
+        query_string = self._job_selection_query(
+            method=method,
+            filter=filter,
+            state=state,
+            username=username,
+            start=start if isinstance(start, int) else 0,
+            limit=limit if isinstance(limit, int) else 0,
+            ascending="True" if ascending else None,
+        )
         response = requests.get(f"{self.mcrit_server}/jobs/{query_string}", headers=self.headers)
         if self.raw:
             return response
         data = handle_response(response)
         if data is not None:
             return [Job(job_data, None) for job_data in data]
+
+    def getQueueCount(self, method=None, filter=None, state=None, username=None):
+        """
+        How many jobs getQueueData would list for the same selection - what a paginated listing
+        needs to size itself.
+        Supported by mcritweb API pass-through
+        """
+        query_string = self._job_selection_query(method=method, filter=filter, state=state, username=username)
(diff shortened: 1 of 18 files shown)


danielplohmann (dismissed):
Approved. Replacing the fetch-everything-and-filter-in-Python path with server-side state queries is the right fix, and the risky part - `_STATE_QUERIES` having to mirror `_identifyJobState` exactly - is covered by test_state_queries_agree_with_identify_job_state, which asserts the same set per state, the same counts, and that every document lands in exactly one state. Composed with #163 and #169: 323 passed, ruff + format + ty clean.

danielplohmann (dismissed):
Re-approving: the earlier approval was dismissed with no intervening push.

danielplohmann (dismissed):
Re-approving after the previous two were dismissed. Verified on current main (2ac8d7b): #168 merges cleanly, and the whole scaling batch #194-#200 then merges cleanly on top of it, so landing this first costs nothing downstream. CI green. Unblocks mcritweb#172.

danielplohmann:
@r0ny123 — I approved this three times and you dismissed it three times, so let me stop doing that and ask instead. Sorry for the noise; I should have read the timeline before the third one.

I see #214 now, and I think I understand: it rewrites the same five files this does — `mongoqueue.py`, `LocalQueue.py`, `QueueRemoteCalls.py`, `JobResource.py`, `McritClient.py` — and #210 makes the case for it plainly (25 sample rows costing 50 requests is a much sharper problem than anything this PR fixes).

What I measured, in case it is useful:

| | result |
|---|---|
| #168 onto `main` | clean |
| #214 onto `main` | clean |
| #168 then #214 | **conflicts on all five files** |
| #214 then #168 | **same five files** |
| #168 is an ancestor of #214 | no |
| #194–#200 then #214 | clean |

So it is symmetric — whichever lands first, the other needs real work, not a mechanical resolution. That is your call to make, not mine, and I would rather sequence it the way you intend than pick for you.

**Which would you prefer?**

1. **#214 first**, and this one rebases onto it — the state queries, text filter and `get_job_count` still look worth having, they are just not what #210 is asking for.
2. **#168 first**, and #214 rebases — it is reviewed and green, but that hands you the conflict on work you opened yesterday.
3. **Fold them together** into one change to `get_jobs`, if the two selections really want to be one query builder.
4. Something else — including dropping this one, if #214 subsumes what it was for.

No rush from my side, and nothing here is blocked on it. The only thing waiting is fkie-cad/mcritweb#172, which wants this PR's job-owner half.

For what it is worth, the review stands on the content: `test_state_queries_agree_with_identify_job_state` is exactly the right test for the risky part, asserting the query and the Python classification select the same set per state and that every document lands in exactly one.

r0ny123:
Thanks, and sorry for the dismissals. They weren't about the review. Option 2, the way option 3 describes it: please land #168 first. I'll rebase #214 onto it and turn `sample_ids`/`job_ids` into conditions in `_job_query`, so there's one query builder and `/jobs/count` counts the same selection. #214 is a day old and yours to review fresh anyway, so the conflict should be on my side. I've also fixed #168's description, which still said it was stacked on #161.

danielplohmann (dismissed):
Landing this first as you asked. Re-verified on current main: merges clean, CI green, and the scaling batch #194-#200 still composes on top of it. Thanks for taking the rebase on #214 - folding sample_ids/job_ids into _job_query so there is one query builder and /jobs/count counts the same selection sounds better than either PR alone.

PR #170, opened 2026-09-06
Answer function name searches from the distinct names

r0ny123:
Closes fkie-cad/mcritweb#76 (searching for functions is painfully slow, ~30 s, on larger databases if no results are found).

## Cause

A free-text function search becomes an unanchored, case-insensitive regex on `function_name`. MongoDB cannot bound an index with that, so when nothing matches it examines every function document before it can answer the empty page: seconds on a million functions, the reported 30 s on larger databases. With a non-default sort (`sort_by=num_blocks` etc.) it examines every document even when there are hits, since it cannot stop early.

## Fix

Function names repeat heavily: most are empty, the rest are symbols shared across samples (the live test database holds 9 distinct names for 1158 functions). So the set of distinct names is small where the collection is large.

`MongoDbStorage.findFunctionByString` now:

1. Lists the distinct names with one `$group` + `$limit` aggregation, which MongoDB runs as a `DISTINCT_SCAN` over the existing `function_name` index (tens of milliseconds for millions of functions, verified with `explain()`). Only when the query actually holds a substring condition on `function_name`; `sample_id:0` or `function_name:main` (equality) do not pay for it.
2. Evaluates the substring condition against that list in Python with the same escaped, case-insensitive pattern, and hands MongoDB an `$in` (or `$nin` for `!?`) of the matching names, which it answers from the index. Nothing matching becomes `$in: []`, so the empty page costs nothing.
3. Above `_DISTINCT_VALUES_CAP` (10000) distinct names the search keeps the regex, unbounded as before. That is the one case this PR does not speed up; it needs an n-gram index, which is out of scope.

`MongoSearchTranspiler` takes an optional `known_values` map for this; an empty search term keeps the regex (it matches everything), other fields and operators are untouched.

## Measurements

Two million synthetic functions, MongoDB 7, page of 101, from the storage's own `_get_search_query` + `find`:

| collection | search | before | after |
|---|---|---|---|
| 5000 distinct names | no result, default sort | 4.2 s (2M docs) | 22 ms (0 docs) |
| 5000 distinct names | name held by 19 functions | 4.2 s | 2 ms |
| 5000 distinct names | `beacon` (515 names), sorted by num_blocks | 1.5 s | 62 ms |
| 5000 distinct names | `main` (986 names), default sort | 15 ms | 78 ms |
| 5000 distinct names | `e` (4014 names), default sort | 5 ms | 27 ms |
| 1.94M distinct names (above cap) | no result | 4.3 s | 4.1 s (regex fallback, +70 ms listing) |

The common-term default-sort case is the only one that gets slower, by tens of milliseconds, because the regex scan in id order finds its 101 hits almost immediately there.

## Verification

- `tests/testMongoSearchTranspiler.py`: matching values become `$in` / `$nin`, case-insensitive like the regex, metacharacters literal, nothing matching is an empty list, empty term and other fields keep their form.
- `tests/testStorage.py` (Mongo): the search answers the same entries through the rewrite and through the regex fallback (cap patched to 2), combined with other conditions and negation, and the listing runs only for substring conditions.
- Full suite: 214 passed, `ruff check`, `ruff format --check`, `ty check` clean.
- Live: server restarted with the change, `/search/functions` for a no-result term, a name in both cases, `sample_id:0 decrypt`, `sample_id:1 decrypt` (no result), `function_name:!?decrypt`, the empty term and a term with regex metacharacters all answer as before; MCRITweb's function listing, sample page and search page render the same rows.

### DIFF
--- a/mcrit/storage/MongoDbStorage.py
+++ b/mcrit/storage/MongoDbStorage.py
@@ -161,6 +161,9 @@ class MongoDbStorage(StorageInterface):
     # DISTINCT_SCAN over the function_name index, tens of milliseconds even for millions of
     # functions; above the cap the search falls back to the regex, unbounded as before.
     _DISTINCT_VALUES_CAP = 10000
+    # and at most this many bytes of them in total: the matching values go into one $in, which
+    # must stay well inside MongoDB's 16 MiB command limit even when they are long mangled symbols
+    _DISTINCT_VALUES_MAX_BYTES = 1 << 20
 
     _database: Optional["Database"]
 
@@ -1818,9 +1821,18 @@ class MongoDbStorage(StorageInterface):
         return sort_list
 
     def _getDistinctValues(self, collection: str, field: str) -> Optional[List[Any]]:
-        """All distinct values of the field, or None when there are more than _DISTINCT_VALUES_CAP."""
+        """All distinct values of the field, or None when there are more than _DISTINCT_VALUES_CAP
+        of them or they exceed _DISTINCT_VALUES_MAX_BYTES in total."""
         pipeline = [{"$group": {"_id": "$" + field}}, {"$limit": self._DISTINCT_VALUES_CAP + 1}]
-        values = [document["_id"] for document in self._getDb()[collection].aggregate(pipeline)]
+        values = []
+        total_bytes = 0
+        for document in self._getDb()[collection].aggregate(pipeline):
+            value = document["_id"]
+            if isinstance(value, str):
+                total_bytes += len(value.encode("utf-8"))
+                if total_bytes > self._DISTINCT_VALUES_MAX_BYTES:
+                    return None
+            values.append(value)
         if len(values) > self._DISTINCT_VALUES_CAP:
             return None
         return values
--- a/mcrit/storage/MongoDbStorage.py
+++ b/mcrit/storage/MongoDbStorage.py
@@ -24,6 +24,7 @@ from mcrit.index.SearchQueryTree import (
     BaseVisitor,
     FilterSingleElementLists,
     NodeType,
+    NotNode,
     OrNode,
     PropagateNot,
     SearchConditionNode,
@@ -54,6 +55,12 @@ class MongoSearchTranspiler(BaseVisitor):
     """
     Converts a tree to a MongoDB query.
     The input tree MUST NOT contain Not or SearchTerm nodes.
+
+    known_values maps a field to the complete list of distinct values it holds. A substring
+    condition on such a field is evaluated against that list here and emitted as an $in / $nin
+    of the values that match, which MongoDB answers from the index. The unanchored,
+    case-insensitive regex it replaces cannot use index bounds and makes MongoDB examine every
+    document of the collection when nothing matches (fkie-cad/mcritweb#76).
     """
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


danielplohmann:
Reviewed this as part of the batch that merges clean against `f83a345`. The rewrite is right and the bounds are the right bounds — a 10,000-value cap plus a 1 MiB byte cap, falling back to the regex above either, is exactly the shape this needs. One measurement before it goes in, because on a corpus this size the feature never fires and the check itself is not free.

Live instance here: 8,699 samples, **11,636,211 functions**.

```
distinct function_name values:            314,144        (31x _DISTINCT_VALUES_CAP)
capped scan, stops at 10,001:   1,031 ms cold, 939 / 830 ms warm
true distinct count:                       25.9 s
```

So every substring search on `function_name` collects 10,001 groups, returns `None`, and falls back to the regex — about **0.9 s added to the ~30 s** that fkie-cad/mcritweb#76 is about. The docstring's "one DISTINCT_SCAN over the function_name index, tens of milliseconds even for millions of functions" holds at 7,244 samples; at 11.6M functions it is ~0.9 s, and `$group` does not get the index-only DISTINCT_SCAN that `.distinct()` would.

Raising the cap does not rescue it: 314k names is what a `$in` would have to carry, which is past the 16 MiB command limit and well past the 1 MiB you set for good reason.

What would work here is **caching the over-cap verdict** rather than recomputing it per search — remember per (collection, field) that the distinct count exceeded the cap, and invalidate on write to that field. The expensive answer is the *negative* one, it is stable (a corpus does not drop below 10,000 distinct names once past it), and a stale positive is the only risk, which invalidation covers. Small corpora keep the win, large ones pay the ~0.9 s once instead of per query.

If a cache is more machinery than you want in this PR, a cheaper version gets most of it: compare `estimated_document_count()` against the cap first and skip the scan when the collection is orders of magnitude larger — wrong in the rare case of a huge collection with few distinct values, but wrong in the direction of today's behaviour rather than slower than it.

Not a blocker in the sense of being incorrect — the fallback does exactly what it promises. It is just the one PR in the batch that could make this instance measurably slower, and it is cheapest to fix while it is still open. The other seven (#205, #211, #212, #206, #160, #175, #162) are good to go from our side; combined they pass 389 tests, ruff, ruff format and ty on the merged tree.

Separately, #162 is worth a deployment line in its changelog entry: its six new compound indexes on `functions` land on a collection holding 11.6M documents and 2.38 GB of indexes here, and they build on first start after the upgrade.

r0ny123:
Thanks for measuring this on the 11.6M-function instance. The 314,144 distinct names and the ~0.9 s capped scan on every search were exactly what I had missed.

a0b0955 caches the over-cap verdict per (collection, field) on the storage object, as you suggested. I cache only the negative answer, never the names. A corpus that has grown past the cap stays past it, so writes never need to invalidate the cache, and each worker process can safely keep its own copy. If deletions bring a corpus back under the cap, the search simply uses the regex, which is today's behaviour. The verdict expires after an hour (and `clearStorage` drops it), so a corpus that shrank gets the fast path back eventually. On your instance, each process pays the ~0.9 s once an hour instead of on every search. A new test checks that two substring searches on an over-cap collection run the capped scan only once; it failed before the change. I also corrected the docstring and the CHANGELOG entry with your numbers: tens of milliseconds at 5,000 names over two million functions, but ~0.9 s to find out that 11.6M functions have more than 10,000 names. The full suite passes including the mongo tests (359 passed against a local mongod 7.0.14), as do ruff, ruff format and ty.

For #162, 7571116 adds the deployment note to its changelog entry. The six new compound indexes on `functions` build on the first start after upgrading, and on a corpus your size (11.6M function documents, 2.38 GB of indexes) that will take noticeable time before the server is ready.

PR #176, opened 2026-09-06
Clean up the query data nothing refers to, and survive jobs without a result (#68)

### LINKED ISSUE #68: Considerations for DbCleanup
* think of orphan query samples and functions that don't have a job connected to them
* think of doing a DB compact afterwards

r0ny123:
Closes #68 (considerations for DbCleanup: orphan query samples and functions without a job, a compact afterwards).

## What the cleanup did, and the bug in it

`doDbCleanup` deleted query samples and query jobs older than `STORAGE_MONGODB_CLEANUP_TTL`, which already covers query samples whose job is gone (they age out by their own timestamp). It read `result["info"]["sample"]` off every finished *and failed* query job, though, and a failed job that died before matching has no result, so the first such job took the whole cleanup down with a `TypeError`. On an instance with `STORAGE_MONGODB_ENABLE_CLEANUP` that means the cleanup never completes again once one query job has failed early.

## Changes

- A job without a result is nothing to protect and nothing to collect: it is deleted once older than the TTL and kept otherwise. Finished jobs are dated by `finished_at`, failed ones by `started_at`, a job that never started counts as old.
- After the TTL pass, `deleteOrphanedQueryData()` on the storage removes the query functions whose query sample is gone and the query disassembly (`query_xcfg`) whose function is gone: what an interrupted deletion or a job deleted without its sample leaves behind. The query collections are bounded by the TTL, so listing their ids is affordable; the memory backend drops orphaned query functions the same way.
- `STORAGE_MONGODB_COMPACT_AFTER_CLEANUP` (default off) runs MongoDB's `compact` on `query_samples`, `query_functions`, `query_xcfg`, `fs.files` and `fs.chunks` afterwards, so the freed space goes back to the file system. It needs the compact privilege; a refusal is reported per collection, not raised, since the cleanup itself has succeeded by then.
- The job now answers a report: samples and jobs deleted, orphans removed, and the compact outcome when it ran.

## Verification

- `tests/testWorkerCleanup.py` (3 tests, worker with fakes): an old failed job without a result is deleted and a recent one kept, a job that never started counts as old; a recent job protects its sample while an old job takes its old sample along and a recent sample is kept; orphan removal always runs and compaction only when configured, with both reflected in the report.
- `tests/testStorage.py` (Mongo): orphaned query functions and disassembly are deleted while the query sample's own are kept, and a second run finds nothing; compact answers per collection against the real MongoDB.
- Full suite: 208 passed; `ruff check`, `ruff format --check`, `ty check` clean.

### DIFF
--- a/docs/TUNING.md
+++ b/docs/TUNING.md
@@ -92,3 +92,21 @@ Notes on the individual knobs:
   earlier only the batch size, `BAND_MATCHES_REQUIRED` and the mongod cache size apply.
 * All measurements used single-process matching; comparisons against a pooled configuration
   will differ.
+
+## Reclaiming space after the query cleanup
+
+| setting | default | effect |
+|---|---|---|
+| `STORAGE_MONGODB_COMPACT_AFTER_CLEANUP` | `False` | run MongoDB's `compact` on `query_samples`, `query_functions` and `query_xcfg` after every `DbCleanup` job |
+
+The cleanup job deletes expired query samples, their functions and disassembly, and the
+orphans a broken deletion left behind; WiredTiger keeps the freed pages inside the collection
+files and reuses them for later inserts, so disk usage does not shrink on its own. `compact`
+returns that space to the file system. It needs the `compact` privilege on the database
+(the default `readWrite` role does not carry it - grant `dbAdmin` or a custom role), it
+blocks writes to the collection it is working on for the duration (seconds to minutes,
+depending on collection size; queries keep being accepted by the other collections), and on
+a replica set it runs on the member it is sent to only. Leave it off unless the query
+collections are large and the instance's disk is tight; the cleanup report says how many
+bytes each compaction returned.
+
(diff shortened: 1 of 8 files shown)


r0ny123:
Live verification on the MongoDB-backed instance (server and worker restarted on this change):

Planted the residue the issue is about: two query functions whose query sample does not exist, two query disassembly documents (one for an orphan function, one for no function at all), and a failed `getMatchesForUnmappedBinary` job from nine days ago with `attempts_left: 0` and no result, which is the shape that used to take the cleanup down.

Scheduled `doDbCleanup` through the index. The job finished and answered:

```
{'num_query_samples_deleted': 0, 'num_query_jobs_deleted': 1, 'orphans': {'query_functions': 2, 'query_xcfg': 2}}
```

Afterwards `query_functions` and `query_xcfg` were empty, the failed job was gone, and the four corpus samples with their 1158 functions were untouched.

PR #177, opened 2026-09-06
Let a family carry the actors it is attributed to (#57)

### LINKED ISSUE #57: Consider adding actors field to families

r0ny123:
Closes #57 (consider adding actors field to families).

## Changes

- `FamilyEntry` carries `actors`, a list of names, in `toDict()`/`fromDict()`; families stored before this read as an empty list, so no migration. `FamilyEntry.normalizeActors()` strips, drops empties and duplicates, keeps the order given.
- Both storage backends accept `actors` in `modifyFamily()`; the MongoDB one sets the field on the family document.
- `PUT /families/{id}` accepts `actors` as a list of names (1 to 64 characters of letters, digits, spaces, dots, dashes, underscores; a comma separated string is accepted too); anything else is a 400 without touching the family.
- `McritClient.modifyFamily(family_id, actors=[...])`; an empty list clears them. The client now sends the update as JSON, since a list does not survive form encoding; the route reads `req.media` and takes both.
- Exports carry `family_actors` (family id to names, only for families that have any); imports merge them into what the target instance already knows for the remapped family. An importer without the key ignores it, an export without the key imports as before.

## Verification

- `tests/testFamilyActors.py` (5 tests): entry round trip and legacy default; normalisation; route validation and pass-through, including the comma separated form and two rejected shapes; the client's JSON body with and without actors.
- `tests/testStorage.py` (both backends): set, normalised, kept through a rename, cleared, unknown family refused.
- `tests/testMinHashIndex.py`: an export carries the actors and an import through JSON merges them with the target's own.

### DIFF
--- a/mcrit/client/McritClient.py
+++ b/mcrit/client/McritClient.py
@@ -154,13 +154,22 @@ class McritClient:
     ### Families
     ###########################################
 
-    def modifyFamily(self, family_id, family_name=None, is_library=None):
+    def modifyFamily(self, family_id, family_name=None, is_library=None, actors=None):
+        """
+        Modify a family: rename it, mark it as library, or set the actors it is attributed to (#57)
+        <actors> is a list of names; an empty list clears them
+        """
         update_dict = {}
         if family_name is not None:
             update_dict["family_name"] = family_name
         if is_library is not None:
             update_dict["is_library"] = is_library
-        response = requests.put(f"{self.mcrit_server}/families/{family_id}", update_dict, headers=self.headers)
+        if actors is not None:
+            update_dict["actors"] = list(actors)
+        # JSON, since a list of actors does not survive form encoding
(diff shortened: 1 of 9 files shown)


r0ny123:
Live verification on the MongoDB-backed instance (server and worker restarted on this change, through `McritClient`):

- `modifyFamily(1, actors=[" Actor One", "Actor Two", "Actor One"])` answered "Family modified."; `getFamily(1)` then carried `['Actor One', 'Actor Two']`, and the family document held the same.
- `modifyFamily(1, actors=["<x>"])` was rejected (400) and the family stayed as it was.
- `getExportData()` carried `family_actors: {'1': ['Actor One', 'Actor Two']}`.
- `modifyFamily(1, actors=[])` cleared them; `getFamily(1).actors == []`.

PR #178, opened 2026-09-06
Optionally keep the binary a sample was submitted as (#95)

### LINKED ISSUE #95: Option to store submitted binaries in MCRIT
Right now, submitted binaries are processed internally via SMDA, the data is translated into MCRIT's data storage and the original binary is most likely discarded or only reconstructable from the respective submission job's data.  
This should possibly change to (configurable, default-disabled) saving of the raw submitted binaries directly in GridFS for future use.

r0ny123:
Closes #95 (option to store submitted binaries in MCRIT).

## Changes

- `STORAGE_KEEP_SUBMITTED_BINARIES` (default off): when set, `Worker.addBinarySample` stores the raw bytes after the sample is indexed and reports `binary_stored` in the job result. Nothing changes for SMDA report submissions, which carry no binary.
- Storage: `storeSampleBinary()`, `getSampleBinary()`, `deleteSampleBinary()` on the interface. The MongoDB backend keeps them in a GridFS bucket `sample_binaries` of the storage database, keyed by `metadata.sample_id` (indexed), replaces on re-store, removes them with `deleteSample` and drops the bucket with `clearStorage`. The memory backend keeps them in a dict.
- `GET /samples/{id}/binary` answers the bytes as `application/octet-stream`, 404 for an unknown sample or one without a stored binary; `McritClient.getSampleBinary()` answers bytes or None.

## Verification

- `tests/testSampleBinaries.py` (4 tests): the route serves octets and 404s both ways; the client answers bytes or None; the worker stores the binary for the indexed sample only when configured.
- `tests/testStorage.py` (both backends): store, read back, replace, delete, and the binary goes with its sample.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,26 @@ reasoning is still at hand, rather than reconstructing it from the commit log at
 
 ## [Unreleased]
 
+### Added
+
+- `STORAGE_KEEP_SUBMITTED_BINARIES` keeps the raw binary a sample was submitted as, in a
+  GridFS bucket beside the corpus, and `GET /samples/{sample_id}/binary` serves it back
+  (`McritClient.getSampleBinary`). Off by default: it stores the sample a second time, so a
+  corpus that turns it on grows by the size of its submissions. A sample submitted before it
+  was turned on has no binary until it is submitted again, which the existing-sample path now
+  completes instead of skipping. Deleting a sample deletes its binary with it.
+- `StorageInterface.hasSampleBinary()` and `.openSampleBinary()`, so neither asking whether a
+  binary is stored nor serving one has to read the file. `openSampleBinary()` answers with a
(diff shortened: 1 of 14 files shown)


r0ny123:
Live verification on the MongoDB-backed instance with `STORAGE_KEEP_SUBMITTED_BINARIES = True` (server and worker restarted on this change):

- Submitted `/usr/bin/cat` (39,384 bytes) through `McritClient.addBinarySample`. The job result carried `binary_stored: True` for the new sample, and one file sat in the `sample_binaries` bucket for it.
- `getSampleBinary(sample_id)` answered bytes equal to the submitted file.
- `getSampleBinary(0)`, a sample that was submitted as an SMDA report, answered `None` (404).
- `deleteSample(sample_id)` removed the sample and its bucket file with it (0 files afterwards).

PR #179, opened 2026-09-06
Keep enough of the SMDA report to rebuild it from storage (#94)

### LINKED ISSUE #94: Ensure SmdaReport can be rebuilt from MCRIT's data storage
Right now, SMDA reports are parsed and translated into MCRIT's database.  
It is likely that some information is lost on the way, i.e. fields being stripped that are not relevant to MCRIT's operation.  
This should be changed in a way that SMDA reports can be reassembled from the data stored in MCRIT, as this would enable further processing options benefiting from the full data view of the original SMDA report.  
To achieve this, all except the xcfg field should be stored in a SampleEntry, and the full function dict should be stored in the FunctionEntry.

r0ny123:
Closes #94 (ensure SmdaReport can be rebuilt from MCRIT's data storage).

## Changes

- `SampleEntry` keeps `smda_extras`: every top-level field of the report's `toDict()` that the entry does not hold in a field of its own, minus `xcfg`, plus the metadata entries it does not hold (`is_buffer`, `language`, ...). It travels in the entry's dict, so the sample document, exports and the API carry it; entries stored before read as `{}`.
- `SampleEntry.toSmdaReportDict(xcfg)` assembles the report dict from the entry's own fields, the extras and the functions' disassembly by offset, with an empty report's defaults for what an older entry did not record.
- `StorageInterface.getSmdaReportForSample()` rebuilds the `SmdaReport` from the entry and `getFunctionsBySampleId(..., with_xcfg=True)` (a new flag, one batched blob fetch on MongoDB); functions whose disassembly was dropped are absent from the xcfg.
- `GET /samples/{id}/smda` and `McritClient.getSmdaReportForSample()` serve it.

## Verification

- `tests/testSmdaReportRebuild.py` (4 tests): the example report round-trips through a `SampleEntry` byte for byte (`report.toDict() == rebuilt.toDict()`), and through the entry's wire dict; an entry stored before the extras still rebuilds; the route and the client.
- `tests/testStorage.py` (both backends): a report added to storage is rebuilt equal to the original; an unknown sample answers None.

### DIFF
--- a/mcrit/client/McritClient.py
+++ b/mcrit/client/McritClient.py
@@ -259,6 +259,18 @@ class McritClient:
         if data is not None:
             return SampleEntry.fromDict(data)
 
+    def getSmdaReportForSample(self, sample_id):
+        """
+        The SMDA report the sample was submitted as, rebuilt from what the server stores (#94); None for an unknown sample
+        """
+        response = requests.get(f"{self.mcrit_server}/samples/{sample_id}/smda", headers=self.headers)
+        if self.raw:
+            return response
+        data = handle_response(response)
+        if data is None:
+            return None
+        return SmdaReport.fromDict(data)
+
     def getSamples(self, start=0, limit=0):
         """
         Get all SampleEntries, optionally from sample_id <start> and up to <limit> many
--- a/mcrit/index/MinHashIndex.py
+++ b/mcrit/index/MinHashIndex.py
@@ -449,6 +449,9 @@ class MinHashIndex(QueueRemoteCaller(Worker)):
     def getFunctionsBySampleId(self, sample_id):
         return self.getStorage().getFunctionsBySampleId(sample_id)
 
+    def getSmdaReportForSample(self, sample_id):
+        return self.getStorage().getSmdaReportForSample(sample_id)
+
     def isFunctionId(self, function_id):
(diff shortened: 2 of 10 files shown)


r0ny123:
Live verification on the MongoDB-backed instance (server and worker restarted on this change), and one thing it caught.

**Caught and fixed before this PR's final state:** the first version kept the extras exactly as `SmdaReport.toDict()` gives them, and the data references there are keyed by integer addresses. MongoDB refuses non-string keys, so on the live instance every binary submission failed at the sample insert (`bson.errors.InvalidDocument: documents must have only string keys, key was 10252`). The example report the unit tests use has no data references, which is why they did not see it. The extras are now stored JSON-shaped (string keys, as a report file holds them), which `SmdaReport.fromDict` reads back the same way it reads a file; the round-trip tests now set data references on the report first, and the amended commit is what this PR carries.

**Verified live afterwards:**
- Submitted `/usr/bin/cat` as a binary; `getSmdaReportForSample(sample_id)` answered a `SmdaReport` with the submission's sha256, 139 functions in its xcfg for 139 stored functions, `smda_version 4.5.0`, family `coreutils`, status `ok`.
- A sample stored before this change (no extras) rebuilt too: 400 functions, defaults for what was not recorded.

PR #181, opened 2026-09-06
Store pichashes zero-padded so that hex order is numeric order (#145)

### LINKED ISSUE #145: Store pichash zero-padded so that hex order is numeric order
## Problem

`pichash` is stored hex-encoded in `functions._pichash`, produced by `hex(value)` in `MongoDbStorage._encodePichash()`. That representation is variable width and unpadded (`0x4d2`, `0x12345678`, `0xffffffffffffffff`), so string order is not numeric order: `"0x9"` sorts after `"0x10"`. Two user-visible consequences, both of which v1.7.1 had to close off rather than fix:

- **Range conditions are not answerable.** `pichash:<0x99` can only be served by comparing the stored strings, which returns a wrong set rather than an empty one. v1.7.1 rejects `<`, `<=`, `>` and `>=` on `pichash` outright (#100).
- **Sorting by `pichash` cannot work.** The search cursor pages by comparing the sort field with a range operator, so the second page of a pichash-sorted function search hit exactly that rejection; and the sort itself named `pichash` while the document stores `_pichash`, so mongod sorted on a field that is not there and the first page was quietly mis-ordered. v1.7.1 rejects the sort on the MongoDB backend with a message that names the field. `MemoryStorage` still sorts by pichash, because it keeps the value as an `int` - so the two backends now answer that request differently.

## Proposal

Store the pichash as fixed-width zero-padded hex - 16 hex digits for a 64-bit value, e.g. `0x00000000000004d2` - so that lexicographic order is numeric order. Then:

- the four range operators can be served directly, with no `$expr` or server-side JS conversion,
- the cursor can page by pichash, so `MongoDbStorage._UNSORTABLE_FIELDS` and the guard around it can go away,
- `_get_sort_list_from_cursor()` still has to map `pichash` to `_pichash`, otherwise the sort keeps naming a field that does not exist,
- `?`/`!?` substring searches keep working, but a term would need to be padded to match at the front - worth deciding whether to left-pad the search term or to document the anchoring.

`_picblockhashes[].hash` is encoded the same way and has the same problem, so it should be padded in the same pass if block hashes are ever to be range-searched or sorted.

## Migration

This changes stored values, so it needs migration tooling in the shape of the v1.7.0 `_xcfg` split: rewrite `_pichash` for every function document with `copy`/`verify`/`inplace`/`unsplit` modes, and a read path that accepts both widths while the migration is in flight. Comparisons must stay rejected until every document is padded - a half-migrated corpus would answer range queries with a plausible wrong set instead of an honest error, which is worse than what v1.7.1 does today. The `_pichash` index is rebuilt as a side effect, and any export/import that carries the encoded form has to be considered.

Not a patch-release change, hence this issue: it gives the v1.7.1 rejections somewhere to point, and it is the only route to numeric range search and sorting on this field.

r0ny123:
Closes #145.

`_pichash` and `_picblockhashes[].hash` hold 64 bit values as variable-width hex strings (`hex()`), so `pichash:<0x99` answered a plausible but wrong set and a pichash-sorted function search could not page: the search cursor compares the sort field with a range operator, which the transpiler rejected for exactly that reason. Written 16 digits wide, string order is numeric order.

## What changes

- **Instance flag** `settings.pichash_padded`. Fresh instances get it at creation; it is reported through `getStats()` and `/status`. The storage reads it once per process.
- **Width-aware encoding.** `_encodePichash(..., padded=)`; `_encodeFunction` and `recalculateAllPicHashes` write in the instance's width. Import goes through `FunctionEntry` integers, so it follows the instance too.
- **Range and sort by pichash** are allowed only on a padded instance (sort key mapped to `_pichash`); an unpadded one rejects both with a message naming the migration, instead of answering a wrong set.
- **Dual-width reads while unpadded.** Search conditions, `isPicHash`, `getMatchesForPicHash` and `getMatchesForPicBlockHash` match both encodings (`$in` on the indexed field), so a migration in flight misses nothing. On a padded instance the point lookups are single-value again.
- **Migration** `python -m mcrit.migrations.migrate_pichash_padding --mode pad|verify|unpad`: resumable keyset walk over `functions` and `query_functions` with positional `$set` per changed value, a leftover sweep for documents a still-running server wrote below the cursor, then the flag. `verify` counts unpadded values and exits 1 when the flag says padded but some remain (a 16-digit value is legitimate under either encoding, so only that direction is a problem). `unpad` rolls back.

Operationally: stop server and workers, run `pad`, restart. Range/sort by pichash on an unmigrated instance keep failing the same way as before, just with a clearer message; everything else behaves as it did.

## Verification

- `tests/testPichashPadding.py` (MongoDB): fresh instance is padded; lookups by value on both shapes; a half-migrated instance misses nothing and still refuses range/sort; range conditions answer the numeric set (`pichash:<0x99` is empty); a pichash-sorted function search pages forward and backward in numeric order; migration round trip, idempotent re-run, leftover detection and CLI exit codes.
- `tests/testMongoSearchTranspiler.py` covers both widths and the sort helpers.
- Full suite: 210 passed. ruff and ty clean.
- Live migration of the running instance follows in a comment.

### DIFF
--- a/mcrit/index/MinHashIndex.py
+++ b/mcrit/index/MinHashIndex.py
@@ -503,6 +503,9 @@ class MinHashIndex(QueueRemoteCaller(Worker)):
                 "num_families": storage_stats["num_families"],
                 "num_functions": storage_stats["num_functions"],
                 "num_pichashes": storage_stats["num_pichashes"],
+                # False on an instance created before pichashes were stored zero-padded and not yet
+                # migrated (migrate_pichash_padding); range conditions and sorting by pichash need True (#145)
+                "pichash_padded": storage_stats.get("pichash_padded", True),
                 # minhashes are only comparable across functions escaped by the same smda; see
                 # mcrit/minhash/EscaperFingerprint.py for why this is worth surfacing
                 "smda_version": SMDA_VERSION,
--- a/mcrit/migrations/migrate_pichash_padding.py
+++ b/mcrit/migrations/migrate_pichash_padding.py
@@ -1,0 +1,204 @@
+#!/usr/bin/env python3
+"""Store pichashes zero-padded so that their hex order is their numeric order (#145).
+
+`functions._pichash` and `functions._picblockhashes[].hash` (and the same fields of
+`query_functions`) hold 64 bit values as hex strings. Written with `hex()`, they have a
+variable width ("0x4d2"), so a range condition or a sort on them orders "0x99" after
+"0x1000". Padding every value to 16 digits ("0x00000000000004d2") makes string order numeric
+order; the storage then allows range operators and sorting by pichash, which the search
+cursor needs to page a pichash-sorted function search.
+
+Modes:
+  pad     - the migration: rewrite every unpadded value in place, batched by function_id
+            (resumable: progress is a keyset cursor in a state document, a killed run costs
+            one batch). A final sweep catches documents the walk did not see, then the
+            settings flag `pichash_padded` is set, which switches every reader and writer
+            of the instance to the padded shape.
+  verify  - count the values that are still unpadded and check the flag against them;
+            exits 1 when the flag says padded but unpadded values remain.
+  unpad   - rollback: rewrite every value with `hex()` again and clear the flag.
+
+Stop the mcrit server and its workers before `pad`: while the flag is unset they keep
+writing unpadded values, which `pad` would have to be re-run for (it is idempotent, and
+`verify` reports leftovers). Storage reads the flag once per process, so restart them
+afterwards. Until the flag is set, all readers accept both widths, so the instance keeps
+answering correctly during the walk; only range and sort by pichash stay rejected.
+
+Usage:
+    python -m mcrit.migrations.migrate_pichash_padding --mode pad
+    python -m mcrit.migrations.migrate_pichash_padding --mode verify
+    python -m mcrit.migrations.migrate_pichash_padding --mode unpad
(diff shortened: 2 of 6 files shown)


r0ny123:
Live migration of a running MongoDB 7 instance (4 samples, 1158 functions) that was created before this change:

**Before** (legacy shape): `/status` had no `pichash_padded`, stored `_pichash` widths were 16/17/18 characters (50 of 1158 short), `pichash:<0x99` answered HTTP 400 "Range operators are not supported", sorting by pichash answered 400, and matching sample 1 gave a pichash aggregation of 162 own / 166 foreign functions matched, 28 self matches.

**Migration**: server and worker stopped, `--mode verify` reported `pichash_padded: false`, 50 unpadded pichashes and 151 documents with unpadded block hashes, no problems. `--mode pad --batch 500` rewrote 192 documents in 0.13 s, swept 0, set the flag; its verify pass reported 0 unpadded values. Stack restarted on the branch.

**After**: `/status` reports `pichash_padded: true`, every `_pichash` is 18 characters, lookups by value still find the same 4 functions for a known pichash, the matching aggregation for sample 1 is identical to the one before the migration, `pichash:<0x99` answers an empty set, a pichash-sorted function search walks all 1158 functions over 24 pages in exactly the numeric order of the collection, `pichash:<pivot` answers exactly the 99 functions numerically below the pivot, and `pichash:>=0x0` paged through the whole corpus.

PR #182, opened 2026-09-06
Make MatchingResult (de)serialisation cheaper and its wire format round-trip (#44)

### LINKED ISSUE #44: Consider use of JSON marshalling accelerators
By transforming additional DTOs into full python dataclasses, it would likely become possible to use an acceleration library like [mashumaro](https://github.com/Fatal1ty/mashumaro) for the (un)marshalling of MatchingResult, which is very expensive for large reports right now.

r0ny123:
Closes #44, which asks to consider dataclass conversion plus a marshalling accelerator for `MatchingResult`. This PR measures where the cost actually is, takes the in-tree gains, and fixes a round-trip bug found on the way.

## Measurement

Synthetic result of 3000 own functions x 80 matches = 240,000 pairs (7.3 MB JSON), best of three, Python 3.11:

| step | before | after |
|---|---|---|
| `json.loads` | 425 ms | (unchanged) |
| `MatchingResult.fromDict` | 411 ms | 311 ms |
| `MatchingResult.toDict` | 327 ms | 268 ms |
| first `filtered_function_matches` access | 532 ms | 395 ms |
| retained memory after `fromDict` | 52.0 MB (217 B/pair) | 34.8 MB (145 B/pair) |
| constructing 240k `MatchedFunctionEntry` alone | 386 ms | 301 ms |

What remains is the materialisation of one Python object per matched pair (about 1.25 µs each, versus 50 ms for 240k plain tuples). A marshalling library such as mashumaro generates the same kind of per-object constructor and would not remove that; the lever beyond this PR is representational (not materialising every pair, e.g. the existing `compact=True` result variant), which is a design change rather than an accelerator.

## Changes

- `MatchedFunctionEntry` uses `__slots__` and keeps the match flags as the integer the wire format carries; `match_is_minhash/pichash/library` are read-only views on it. That also makes `getMatchTuple` hand the flags back unchanged (the #155 fix, expressed by construction).
- `fromDict`: local append, set-backed membership tests instead of list scans, in-place sort.
- **Bug fix:** `toDict` emitted `matches.functions` as a dict keyed by function id, while the matchers produce (and `fromDict` reads) a list of per-function summaries, so `fromDict(toDict())` raised `TypeError`. It now emits the list, grouped from the sorted `function_matches`.
- Test: round trip through the wire format equals the example matching report, plus the flag views.

Full suite: 204 passed; ruff and ty clean.

### DIFF
--- a/mcrit/storage/MatchedFunctionEntry.py
+++ b/mcrit/storage/MatchedFunctionEntry.py
@@ -1,16 +1,35 @@
-from typing import TYPE_CHECKING, List, Optional
+from typing import List, Optional
 
 import mcrit.matchers.MatcherFlags as MatcherFlags
 
-if TYPE_CHECKING:  # pragma: no cover
-    pass
-
-# Dataclass, post init
-# constructor -> .fromSmdaFunction
-# assume sample_entry, smda_function always available
+_ALL_MATCH_FLAGS = MatcherFlags.IS_MINHASH_FLAG | MatcherFlags.IS_PICHASH_FLAG | MatcherFlags.IS_LIBRARY_FLAG
 
 
 class MatchedFunctionEntry:
+    """One (own function, matched function) pair of a MatchingResult.
+
+    A large result holds hundreds of thousands of these, so the class is kept lean (#44):
+    slots instead of a per-instance dict, and the match flags are kept as the integer the
+    wire format carries, with the match_is_* booleans derived from it on access. That also
+    makes the round trip exact by construction: getMatchTuple hands the flags back as they
+    came in (#155).
+    """
+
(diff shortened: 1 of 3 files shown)


r0ny123:
Live check against the running MongoDB instance with this branch deployed to server and worker: a fresh matching job for sample 2 (force_recalculation) parsed into 11 pairs with flags 1 and 3 (`flags(mp.)` on the first entry), `toDict()` produced the list shape, the per-function summaries equal the worker's own (the worker's carry an extra `num_instructions`, which the entry never held), `fromDict(toDict())` reproduced identical match tuples, and the link hunt on the parsed result still runs.

Issue #158, opened 2026-08-31
getSampleSearchResults 500s on any sha256 that is not in the collection


r0ny123:
Any search for a 64-hex string that isn't a sample in the collection is a 500.

## The line

`mcrit/index/MinHashIndex.py`, in `getSampleSearchResults`:

```python
        if re.match("^[a-fA-F0-9]{64}$", search_term) is not None:
            sample_entry = storage.getSampleBySha256(search_term)
            sha_match = sample_entry.toDict()
        else:
            sha_match = None
```

`getSampleBySha256` returns `None` for a hash that isn't there — both storages do, explicitly:

- `MemoryStorage.getSampleBySha256` → `Optional["SampleEntry"]`, returns `None` when the sha isn't in `_sample_by_sha256`
- `MongoDbStorage.getSampleBySha256` → `find_one(...)`, `target_sample = None` unless a document comes back

so `.toDict()` on the next line raises `AttributeError`.

Note the `try: ... except Exception: pass` just above **closes before this block** — it wraps only the integer-id branch, so it does not cover this.

And `StatusResource._respond_search` catches only `ValueError`, so the `AttributeError` escapes as a 500.

## To reproduce

Search the samples collection for any well-formed sha256 you haven't imported — e.g. 64 `a`s. Pasting a hash you're *checking for* is the obvious thing to do with that box, so this fires on the most natural use of the feature.

## Fix

```python
sample_entry = storage.getSampleBySha256(search_term)
sha_match = sample_entry.toDict() if sample_entry is not None else None
```

`getFamilySearchResults` and `getFunctionSearchResults` are worth a look at the same time — I only traced the sample one, but the id-match branches have the same shape.

Context: this is the backend half of a MCRITweb issue about pasting a sha256 into search. MCRITweb now works around it by asking the backend a second, narrower question when the search fails, but the 500 is here.

Issue #157, opened 2026-08-31
getUniqueFamilyScoreForSample keeps the last match's score, not the best


r0ny123:
Small one, but it's rendered on every 1-vs-N result page, so people are looking at it.

## The line

`mcrit/storage/MatchingResult.py:373`, inside the loop over `self.function_matches`:

```python
weighted_bytes_per_function_id[function_match_summary.function_id] = (
    function_match_summary.num_bytes * function_match_summary.matched_score / 100.0
)
```

Plain assignment. The two collections beside it accumulate — `families_matched_by_function_id[...].add(...)` and `samples_matched_by_function_id[...].add(...)` — so the loop is genuinely per *match*, and a function that matched several times in the same family keeps whichever match happened to be iterated **last**, not the best one.

The `= 0` initialiser a few lines up makes it read as if it were accumulating, which is probably why it survived.

## What it costs

`weighted_bytes_per_function_id` feeds `unique_family_scores_per_sample[...]["bytes_matched"]`, which becomes `unique_score`. So the "unique" percentage on a 1-vs-N result is computed from an arbitrary match rather than the strongest one, and it can only ever be too low.

Measured over the reports I had captured: **81 of 324** unique-family functions in one report (and 28 of 114 in another) have matches with differing scores, so for those the value depends on iteration order. The worst gap I saw between the last-iterated score and the best available was **43.75 points**.

## Fix

```python
weighted = function_match_summary.num_bytes * function_match_summary.matched_score / 100.0
weighted_bytes_per_function_id[fid] = max(weighted_bytes_per_function_id[fid], weighted)
```

`max` rather than sum, since the intent is clearly "this function's contribution", not "add it once per match".

Issue #156, opened 2026-08-31
Score denominator uses config MINHASH_FN_MIN_INS while pichash matching uses the per-request pichash_size


r0ny123:
Another one from auditing MCRITweb against v1.8.1. This one makes a percentage able to exceed 100.

## The mismatch

Two different thresholds decide who gets into the numerator and who gets into the denominator, and only one of them is configuration.

**Denominator** — `MatcherInterface._summarizeMatches`, `mcrit/matchers/MatcherInterface.py:588`:

```python
if function_entry.num_instructions >= self._worker._minhash_config.MINHASH_FN_MIN_INS:
```

That is read from config.

**Numerator** — pichash matching admits a function at `mcrit/matchers/MatcherInterface.py:433`:

```python
if function_size < self._pichash_size:
```

and `self._pichash_size` is set from a **per-request query parameter** (`MatcherInterface.py:82`, filled by `mcrit/server/utils.py:21-24`, which clamps it only to `>= 0`).

## Why that bites

With the default `MINHASH_FN_MIN_INS = 10`, a request carrying `?pichash_size=1` puts every pichash-matched function of 1–9 instructions into the numerator while the denominator excludes all of them. `matched.percent.*` can then come out above 100.

It is reachable without touching any config: MCRITweb's API proxy forwards `pichash_size` straight through on `/api/query/binary` and `/api/query/binary/mapped/<addr>`, so a caller sets it per request.

Separately, and at any configuration, there is an off-by-one between the two comparisons even when the numbers agree:

- denominator: `num_instructions >= MINHASH_FN_MIN_INS`
- `MinHasher.isMinHashableFunction`: `num_instructions > MINHASH_FN_MIN_INS` (`mcrit/minhash/MinHasher.py:51` — and `MINHASH_FN_MIN_BLOCKS` defaults to `0`, so the `or` branch never fires)

so a function of exactly `MINHASH_FN_MIN_INS` instructions sits in the denominator but can only ever be matched by pichash. On the corpus I had to hand that is 18 functions, 0.34% of matchable bytes — small, but it is a systematic bias rather than noise.

## Suggestion

Derive the denominator from the *effective* threshold the matcher actually used for this request rather than from config, or refuse a `pichash_size` below `MINHASH_FN_MIN_INS` at the request boundary. And pick one of `>` / `>=` for the two size tests.

I have not measured a real over-100 score against a live instance — this is read from the source plus the corpus arithmetic — so worth a sanity check before acting on it.

Issue #152, opened 2026-08-30
Serving a matching report parses the stored JSON and re-serialises it: 0.66s of a 1.03s report open


danielplohmann:
Serving a matching report parses the stored JSON and immediately re-serialises it. On a real 8 MB report that round trip is **0.66 s of a 1.03 s report open**, and the payload was already JSON on disk.

### Measured

Live instance, mcrit 1.8.1, 8,691 samples / 11.67M functions. A `getMatchesForSample` report for `win.citadel` (sample 601): **8,078,325 bytes**. Client is a container on the same host as the server, so this is close to best case.

Splitting one `GET /jobs/<job_id>/result`:

| step | |
|---|---|
| time to first byte | **0.699 s** |
| body transfer | **0.012 s** (675 MB/s) |

and inside that server time:

| step | |
|---|---|
| `getResultForJob` — GridFS read + `json.loads` | 0.163 s |
| `jsonify` — re-serialise the dict back to bytes | **0.498 s** |
| transfer | 0.012 s |
| client `json.loads` — parse it a second time | 0.147 s |
| `MatchingResult.fromDict` | 0.176 s |
| **total, first byte to usable object** | **~1.03 s** |

**0.661 s of that 1.03 s is JSON round-tripping.** Nothing consumes the intermediate dict.

### The code

The result is written to GridFS as JSON (`mcrit/libs/mongoqueue.py:471`):

```python
def _dicts_to_grid(self, dicts, **kwargs):
    return self._file_to_grid(json.dumps(dicts).encode("ascii"), **kwargs)
```

and read back by parsing it (`mcrit/libs/mongoqueue.py:474-479`):

```python
def _grid_to_dicts(self, grid, **kwargs):
    result = None
    grid_file = self._grid_to_file(grid, **kwargs)
    if grid_file is not None:
        result = json.loads(grid_file)
    return result
```

`JobResource.on_get_job_result` (`mcrit/server/JobResource.py:136-152`) then turns it straight back into bytes:

```python
data = self.index.getResultForJob(job_id)
if "compact" in req.params and req.params["compact"].lower().strip() == "true":
    ...
    data["matches"].pop("functions")
resp.data = jsonify({"status": "successful", "data": data})
```

So the default path exists only to wrap already-correct bytes in `{"status": "successful", "data": …}`. `on_get_results` (`:115`) has the same shape.

### Suggested fix

[...]

PR #187, opened 2026-09-08
Bump version to 1.9.0 and add the Version History entry

danielplohmann:
Cuts **v1.9.0** from the 13 PRs merged above v1.8.1. Two files change: `pyproject.toml` and `mcrit/config/McritConfig.py` (the version `/status` reports), plus the Version History entry.

### Why 1.9.0, and why the 1.8.2 entry is folded in

`1.8.2` was bumped on the `explicit-band-projection` branch and merged, but **no release was ever cut from it and no `v1.8.2` tag exists** — so leaving it would stand as a phantom release describing one feature out of thirteen. Its changelog text is folded into this entry verbatim.

`1.9.0` rather than a patch bump because the release adds endpoints (`rebuild_picblockhash_index`, `recompute_family_stats`, `repair_minhashes`) and changes the *type* of `MatchedFunctionEntry.match_is_minhash` / `match_is_pichash` / `match_is_library` from the masked integer to a `bool` property. Truthiness is unaffected — `if entry.match_is_pichash` behaves as before — but code comparing them to `2` or `4` has to change. That is called out as BREAKING-for-Python-consumers in the entry.

### What is in it

| | |
|---|---|
| correctness | #155 flag round trip, #156 percentage denominator, #157 best-match weighting, #158 unknown-sha256 search, #44 `fromDict(toDict())` |
| reliability | #150 stranded jobs, #151 family counters, #149 band tombstones |
| operations | #142 selective minhash repair, `getUnhashedFunctions` skipping unhashable functions |
| performance | #152 stored-bytes result serving, `getUniqueBlocks` 92 s → 0.11 s, the dropped `_picblockhashes.offset` index |
| configuration | the `explicit` band projection strategy and its presets |

### Verified against a real corpus, not only the suite

**Matching results are unchanged**, which is the claim the entry leads with, so it was measured rather than assumed. On an 11.6M-function instance (8,691 samples), `win.citadel` at `k=1` under the deployed `segment-balanced-v1` projection:

```
digest : 2ceca3d52f310f84be809182a73b20b1   (identical to the v1.8.1 baseline)
length : 8,125,782 bytes of match payload   (unchanged)
```

The digest covers `match_report["matches"]` in full — 5,523 sample matches plus the function matches.

**The statement is qualified in the entry, deliberately.** #156 and #157 *do* change reported numbers under the right conditions: #156 only when a request passes a `pichash_size` different from the configured `MINHASH_FN_MIN_INS`, and on that instance both are `10`, so it is a no-op there. A caller passing a different `pichash_size` would see different percentages. "Results unchanged" without that caveat would have been the kind of claim that gets discovered as false later.

### Upgrade note

The entry groups the **five one-time operator actions** these PRs introduced into one place, since they arrived across six of them and none is automatic:

* `rebuild_picblockhash_index` — enables the fast `getUniqueBlocks`; until it runs, the old scan is used, which is retained rather than deleted, so upgrading changes performance and never results
* `recompute_family_stats` — corrects counters already drifted
* `repair_minhashes` — rehashes only samples an older escaper hashed
* `purgeEmptyBandDocuments()` — clears band tombstones older deletions left
* `db.functions.dropIndex("_picblockhashes.offset_1")` — reclaims 0.678 GB on an existing instance

No database-shape change, so **no migration is required**.

Also noted: the new endpoints mean MCRITweb needs a matching release to *use* them; MCRITweb works unchanged against this version either way.

```
281 passed, 36 subtests passed
All checks passed!            # ruff check
141 files already formatted   # ruff format --check
All checks passed!            # ty check (pinned 0.0.74)
```

### Not done here

No tag is pushed and no PyPI upload is done — this is the version bump and the entry; cutting the actual release stays a maintainer action.


r0ny123:
@danielplohmann can this changelog be formatted into list, so its get easier to read. Also, can we use this https://github.com/olivierlacan/keep-a-changelog across the mcrit ecosystem?

Issue #186, opened 2026-09-08
generateYaraRule's wrap_at is never read: renderRule tests the wrap_string function instead


danielplohmann:
`renderRule` tests the *function* `wrap_string` for truthiness instead of its `wrap_at` parameter, so the parameter is never read and the wrapping is unconditional.

`mcrit/storage/UniqueBlocksResult.py:63` on `main`:

```python
def renderRule(self, block_cover, condition_required, wrap_at=0):
    ...
            if wrap_string:                       # <- module-level function, always truthy
                yarafied += f"        $blockhash_{pichash} = {{\n"
                yarafied += "            " + wrap_string(result["escaped_sequence"], max_column_length=80, padding=12) + "\n"
                yarafied += "        }\n"
            else:
                yarafied += f"        $blockhash_{pichash} = {{ " + result["escaped_sequence"] + " }\n"
```

`wrap_at` appears exactly three times in the file — the two signatures and the pass-through — and is never read:

```
37:    def generateYaraRule(self, ..., condition_required=7, wrap_at=0):
39:        return self.renderRule(cover, condition_required, wrap_at=wrap_at)
41:    def renderRule(self, block_cover, condition_required, wrap_at=0):
```

Consequences:

1. **`wrap_at` does nothing.** The default is `0`, which reads as "do not wrap", but rules are always wrapped — and at a hardcoded `max_column_length=80`, so a caller asking for a different width silently gets 80.
2. **The `else` branch is unreachable**, so the single-line rule form cannot be produced at all.
3. **A caller is already affected.** MCRITweb's `data.build_yara_rule` calls `ubr.generateYaraRule(wrap_at=40)` (`mcritweb/views/data.py:285`) and gets 80-column wrapping. fkie-cad/mcritweb#140 is adding user-facing rule knobs on top of this, so the knob would arrive dead.

The fix is `if wrap_at:`, presumably passing it through as the column width rather than leaving 80 hardcoded — but which of those two the intended behaviour is (a boolean "wrap or not" plus a fixed width, or a width) is the maintainer's call, which is why this is an issue rather than a PR.

[...]

PR #188, opened 2026-09-08
Adopt Keep a Changelog, keeping the entries' depth

danielplohmann:
Adopts [Keep a Changelog 1.1.0](https://keepachangelog.com/en/1.1.0/), as r0ny123 suggested — with two additions the spec does not cover and this project needs. **One file changes.**

### What the spec gives us

The six standard sections (`Added` / `Changed` / `Deprecated` / `Removed` / `Fixed` / `Security`, omitted when empty), an `Unreleased` section, ISO dates, and comparison links at the bottom.

**The `Unreleased` section is the real win, not the formatting.** The 1.9.0 entry in this PR was reconstructed from 14 merge commits after the fact — which is exactly the "commit log diff" anti-pattern the spec warns about, done by hand. An `Unreleased` section means each change is written down when it merges, while the author still remembers the caveat.

### Two departures from the spec, both deliberate

**A release summary paragraph under the version heading.** *"Matching results are unchanged at default configuration"* is the single most load-bearing sentence in an mcrit release note, and the bullet structure gives a reader nowhere to meet it first.

**An `### Upgrading` block.** 1.9.0 alone introduces five one-time operator actions, arriving across six PRs. The spec has no section for them, and a reader should not have to assemble that list out of the bullets.

### What is *not* converted, and why

**The 125 historical entries stay verbatim**, under an `## Older releases` heading. Translating them is a large rewrite that risks losing nuance in entries whose whole value is their nuance — the v1.7.0 entry alone carries the byte-for-byte verification, the "no migration window needed because readers fall back", the "in-place does not reclaim the space", and the "older versions then fail *quietly*, so run `unsplit` before downgrading". Every version still has an entry, which is what the spec actually asks for.

Verified mechanically rather than eyeballed — exactly one entry left the old list, the one now converted:

```
old: 126 version entries   new: 125
missing from new:  * 2026-09-08 v1.9.0:      <- converted to the new format
added to new:      (none)
```

All twelve link references were checked against the API for the right target type (`issues/` vs `pull/`); #147 and #154 are PRs, the other ten are issues.

### The semver claim is scoped, not asserted flat

The header says the project adheres to semver over **the REST API, the configuration surface and the stored data shape**, and that Python object internals may change in a minor release.

This matters on day one: 1.9.0 changes `MatchedFunctionEntry.match_is_*` from masked integers to bool properties under a minor bump. A bare "adheres to Semantic Versioning" would be false the moment it was written. Worth a maintainer's opinion on whether that is the scope you want to commit to — it is the one judgement in here that is a policy choice rather than a formatting one.

### The convention, stated in the header

> Entries carry the measurement, the caveat and the failure mode, not just the change: a line that says what moved without saying what it cost, what it cannot see, or how it would be noticed is worth less than the space it takes.

The format is a better container than what was there; it will not preserve the depth by itself. That depends on writing the bullets that way, so the expectation is written down where the next person will see it.

### Not done here

- No tooling. The format is machine-parseable enough that the release workflow could assert the top section matches the tag being cut — worth doing, but separately.
- `Deprecated` and `Security` are absent from 1.9.0 because they are empty; the spec expects omission.
- No `CONTRIBUTING.md` — the convention lives in the changelog header, where someone about to write an entry will actually read it. Happy to move it if you would rather it lived elsewhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,142 @@
 # Changelog
 
-Newest first. Entries carry the release date, the version, and what changed.
+All notable changes to this project are documented in this file.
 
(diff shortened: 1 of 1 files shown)


danielplohmann:
Rebased onto `main` and added one line: #186 (the `wrap_at` fix, merged as #189) now appears under `### Fixed` in the 1.9.0 entry, with its link reference.

It goes in **1.9.0 rather than `[Unreleased]`** because the tag has not been cut yet — which is the format working as intended: the entry tracks what the release will contain, rather than being reconstructed from the commit log after the fact.

PR #193, opened 2026-09-12
Release on a tag through trusted publishing, with the checks that were memory before

r0ny123:
Releasing MCRIT is `make publish` with an API token. GitHub releases stopped at v1.3.0, and nothing checks that the three places the version is written — `pyproject.toml`, `McritConfig.VERSION` (served by `/version`) and the newest `CHANGELOG.md` heading — agree before a version number is consumed on PyPI.

This is one of six pull requests giving the MCRIT ecosystem one release process — the same guard, the same tag-triggered workflow, the same changelog check, and a `RELEASING.md` in each repository stating only the values that differ. A maintainer learns it once.

## Architecture

**The hand-written `CHANGELOG.md` is the single stream of change information.** #191 adopted Keep a Changelog; this makes it load-bearing. Every PR that touches `mcrit/` or `pyproject.toml` adds its own bullet under `## [Unreleased]`, or carries the `no-changelog` label; a new `Changelog` check enforces that (read-only token, no PR content executed, Dependabot exempt). At release time the version's section *is* the release notes, and GitHub's generated list of merged PRs and contributors is appended under it. `.github/release.yml` groups the dependency bumps.

Release Please, semantic-release, towncrier and scriv were considered. They either derive notes from commit subjects — which this repository does not write in a machine-readable form, and which yield a list rather than the measured entries the changelog header asks for — or add a second place to write release information. Making the existing file authoritative and gating on it is the cheapest reliable system.

**Versioning** stays Semantic Versioning over the REST API, configuration surface and stored data shape, as the changelog header states. `pyproject.toml` and `McritConfig.VERSION` both stay: the config value is what `/version` serves from a checkout, and reading installed metadata there would break `pip install -e .`-less deployments. A new test asserts the two agree; the workflow refuses a tag that does not match both.

## What pushing `vX.Y.Z` does now

`.github/workflows/publish-release.yml`:

| job | what it does |
|---|---|
| verify | refuses unless the tag matches both version strings, `CHANGELOG.md` has a `## [X.Y.Z] - <date>` section (written out as the notes), the commit is an ancestor of `main`, and the `CI` workflow concluded `success` on that commit |
| build | `python -m build` in isolation, `twine check --strict`, then installs the wheel into a clean venv with only the declared runtime dependencies and imports it with `python -P` so `./mcrit` cannot stand in for it — checks `McritConfig.VERSION`, that `cache/logbuckets.json` shipped, that `shinglers/archived` did not, and runs `mcrit --help` |
| publish | `pypa/gh-action-pypi-publish` through trusted publishing, `id-token: write` as the only credential, attestations on. Environment `pypi`; a manual run goes to `testpypi` and skips an existing version |
| release | `gh release create` with the changelog section as body, `--generate-notes` appended, sdist and wheel attached. Pre-release tags (`v1.10.0rc1`) are marked `--prerelease` and not `--latest` |

A manual run from a tag rehearses the whole path against TestPyPI and stops before the GitHub release. The guard script has tests (`tests/testReleaseGuard.py`).

## Other changes

- **Python floor 3.12** (`requires-python`, classifiers, unit matrix, `ruff` target). The ecosystem shares one floor; the reference `docker-mcrit` deployment already runs 3.12 and nothing in MCRIT needed it. Recorded under `Removed`.
- `pyproject.toml` gains a `Changelog` URL; it was already PEP 621 with a PEP 639 license expression, so nothing else moved.
- `RELEASING.md` documents cutting, rehearsing, pre-releases, recovery, the one-time maintainer configuration, and the release order across the ecosystem (smda → MCRIT → MCRITweb → docker-mcrit). `AGENTS.md` points at it.
- `McritConfig.VERSION` keeps a one-line note that the release workflow checks it against `pyproject.toml`, replacing the note-to-self about the README.

## Security

- Actions were already SHA-pinned with `persist-credentials: false` and per-job permissions; the new workflows follow the same pattern, `contents: read` at the top and raised per job only where a job writes.
- No token stored anywhere — trusted publishing replaces `make publish`'s API token, and each file carries PEP 740 provenance.
- `zizmor --persona=regular` reports no findings; `actionlint` clean.

## Tests

- Unit suite (`-m "not mongo and not sleep"`): 196 passed, 3 failed — the three failures in `testMinHashRepair.py` fail identically on `main` in this environment and are unrelated (they depend on the locally installed smda version). `ruff check`, `ruff format --check` and `ty check` clean.
- `python -m build` + `twine check --strict` pass; the wheel installs into a clean Python 3.13 venv, imports from `site-packages`, ships the cache data and not the archived shinglers, and `mcrit --help` runs. Metadata reads `Requires-Python: >=3.12`, `License-Expression: GPL-3.0-only`.
- `release_guard.py --tag v1.9.0` against the real tree extracts the 1.9.0 section from `CHANGELOG.md`.

## What a maintainer has to configure once

1. A **trusted publisher** on pypi.org for `mcrit`: owner `danielplohmann`, repository `mcrit`, workflow `publish-release.yml`, environment `pypi`; the same on test.pypi.org with environment `testpypi` for rehearsals.
2. GitHub **environments** `pypi` and `testpypi`. A required reviewer on `pypi` turns the publish into an approval step.
3. The **`no-changelog` label**.

Until 1 is done, pushing a tag fails at the publish job; nothing is created on GitHub and the tag can be deleted and re-pushed.

## Compatibility

`make publish` remains the manual fallback. `docker-mcrit` installs `mcrit==<tag>` from PyPI and clones the tag, and both continue to work; the 3.12 floor matches its image. MCRITweb's `mcrit>=1.5.3` floor is untouched.

## Related pull requests across the ecosystem

The same process, one PR per repository. None depends on another to merge; release order once merged is smda → purepdb → picblocks → MCRIT → MCRITweb → docker-mcrit.

- smda: danielplohmann/smda#343
- purepdb: danielplohmann/purepdb#59
- picblocks: danielplohmann/picblocks#5
- MCRIT: danielplohmann/mcrit#193
- MCRITweb: fkie-cad/mcritweb#176
- mcrit-plugin: danielplohmann/mcrit-plugin#15
- docker-mcrit: danielplohmann/docker-mcrit#12

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,24 @@
+# Not a workflow: GitHub reads the generated-release-notes configuration from this path and
+# nowhere else. The release itself is cut by the workflow under `.github/workflows/`.
+#
+# The generated list is appended under the hand-written CHANGELOG.md section, so it exists to
+# credit contributors and link every pull request, not to describe the release. Dependency and
+# tooling bumps are grouped so they do not sit between the real changes.
+changelog:
+  exclude:
+    authors:
+      - github-actions[bot]
+  categories:
+    - title: Changes
+      labels:
+        - "*"
+      exclude:
+        labels:
+          - dependencies
+          - github_actions
+          - pre_commit
+    - title: Dependency and tooling updates
+      labels:
+        - dependencies
+        - github_actions
+        - pre_commit
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,64 @@
+name: Changelog
+
+# pull_request rather than pull_request_target: this needs the PR's changed-file list and no
+# secrets, so the read-only token a fork PR receives is enough, and no PR content is checked
+# out or executed. `labeled`/`unlabeled` are here so applying the skip label re-runs the check
+# rather than leaving a stale failure behind.
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - labeled
+      - unlabeled
+    branches:
+      - main
+
+permissions:
+  contents: read
+  pull-requests: read
+
+jobs:
+  changelog-entry:
+    name: Changelog entry
+    runs-on: ubuntu-latest
+    steps:
+      # The job runs unconditionally and decides inside the step, rather than skipping on an
+      # `if`: a skipped job reports as neither passing nor failing, and the reason a PR is
+      # exempt belongs in the log where the next reader will look for it.
+      - name: Require a changelog entry beside a user-visible change
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPOSITORY: ${{ github.repository }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          # contains() matches an array element exactly, so a label carrying a comma or a
+          # name this one is a prefix of cannot be mistaken for it
+          SKIP_LABELLED: ${{ contains(github.event.pull_request.labels.*.name, 'no-changelog') }}
+          IS_DEPENDABOT: ${{ github.event.pull_request.user.login == 'dependabot[bot]' }}
+        run: |
+          set -euo pipefail
+
+          if [ "${SKIP_LABELLED}" = "true" ]; then
+            echo "Labelled no-changelog; entry not required."
+            exit 0
+          fi
+          if [ "${IS_DEPENDABOT}" = "true" ]; then
+            echo "Dependency bump; the generated release notes list these under their own heading."
+            exit 0
+          fi
+
+          changed="$(gh api --paginate "repos/${REPOSITORY}/pulls/${PR_NUMBER}/files" --jq '.[].filename')"
+
+          if ! printf '%s\n' "${changed}" | grep -qE '^(mcrit/|pyproject.toml$)'; then
+            echo "No shipped files changed; entry not required."
+            exit 0
+          fi
+
+          if printf '%s\n' "${changed}" | grep -qx 'CHANGELOG.md'; then
+            echo "CHANGELOG.md touched beside the change."
+            exit 0
+          fi
+
+          echo "::error file=CHANGELOG.md::This PR changes shipped files but does not touch CHANGELOG.md. Add an entry under ## [Unreleased], or apply the no-changelog label if the change genuinely needs no entry."
+          exit 1
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -1,0 +1,232 @@
+name: Publish release
+
+# Pushing a vX.Y.Z tag is the release. A manual run rehearses the same path against TestPyPI
+# and stops before the GitHub release, so the whole workflow can be exercised without
+# consuming a version number on PyPI.
+on:
+  push:
+    tags:
+      - "v*"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+# A cancelled publish can leave PyPI holding some files of a version and not others, and the
+# version number cannot be reused afterwards.
+concurrency:
+  group: release-${{ github.ref }}
+  cancel-in-progress: false
+
+defaults:
+  run:
+    shell: bash
+
+env:
+  PYTHON_VERSION: "3.12"
+
+jobs:
+  verify:
+    name: Verify the tag
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: read
+      actions: read
+    outputs:
+      version: ${{ steps.guard.outputs.version }}
+      prerelease: ${{ steps.guard.outputs.prerelease }}
+    steps:
+      - name: Refuse anything that is not a tag
+        env:
+          REF_TYPE: ${{ github.ref_type }}
+          REF_NAME: ${{ github.ref_name }}
+        run: |
+          if [ "${REF_TYPE}" != "tag" ]; then
+            echo "::error::Run this from a tag, not ${REF_TYPE} ${REF_NAME}."
+            exit 1
+          fi
+
+      - name: Checkout the tag
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Set up Python
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+
+      - name: Check the tag against the packaged version and the changelog
+        id: guard
+        env:
+          TAG: ${{ github.ref_name }}
+        run: |
+          python .github/workflows/scripts/release_guard.py \
+            --tag "${TAG}" --notes release-notes.md --github-output "${GITHUB_OUTPUT}"
+
+      - name: Check the tag is on main
+        env:
+          TAG: ${{ github.ref_name }}
+        run: |
+          git fetch --quiet origin main
+          if ! git merge-base --is-ancestor "${TAG}" origin/main; then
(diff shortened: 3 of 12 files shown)


danielplohmann:
**Carry the concurrency-group fix into this port before it merges.**

This PR adds:

```yaml
concurrency:
  group: release-${{ github.ref }}
  cancel-in-progress: false
```

A rehearsal (`workflow_dispatch` on a tag) and the real release (`push` of that tag) carry the **same** `github.ref`, so that group holds both. With `cancel-in-progress: false` the rehearsal queues behind the real run — which is parked on the `pypi` reviewer gate, waiting for an approval that the rehearsal was supposed to inform. It never starts.

The fix is one line:

```yaml
  group: release-${{ github.ref }}-${{ github.event_name }}
```

Two real publishes of one tag are both `push`, so they still share a group and still serialise. That is the property the group was added for and it is unaffected.

Fixed upstream in smda danielplohmann/smda#353 and purepdb danielplohmann/purepdb#67. Flagging here so this port does not inherit the shape and need the same fix afterwards.

danielplohmann:
Before this lands, please bring `publish-release.yml` level with smda's current master, so all four PyPI repos run one workflow:

1. **Milestone gate** — the `Check the milestone for this tag is finished` step in `verify` and the `Close the milestone for this tag` step in `release`, byte-identical to smda/purepdb, plus the `issues:` permissions they need.
2. **Concurrency group** — `group: release-${{ github.ref }}-${{ github.event_name }}` with the comment from danielplohmann/smda#353 (merged today). With the ref alone, a TestPyPI rehearsal queues behind a real run parked on the `pypi` reviewer gate.

The rest already matches. `PYTHON_VERSION` can stay as it is here.

You have Write on this repo now, so feel free to move the branch into familiary instead of your fork. main is still PR plus code-owner review.

r0ny123:
@danielplohmann both changes are in as of f90c722e8109bc63638ced94fd5c98b12066ccb9:

1. Milestone gate: "Check the milestone for this tag is finished" in `verify` (with `issues: read`) and "Close the milestone for this tag" in `release` (with `issues: write`), copied from smda master.
2. Concurrency group is now `release-${{ github.ref }}-${{ github.event_name }}`, with the comment from smda#353.

danielplohmann:
Reviewed this and I want it in. I ran the guard rather than reading it — against the real tree, under 3.12 since `tomllib` needs it:

```
v1.9.0     -> matches the declared version and has a changelog section
v1.10.0    -> does not match the packaged version (pyproject.toml = 1.9.0, McritConfig.VERSION = 1.9.0)
1.9.0      -> does not start with 'v'
v1.9       -> is not vMAJOR.MINOR.PATCH with an optional a/b/rc pre-release suffix
--notes    -> extracted the 134-line 1.9.0 section correctly
```

Then rehearsed the cut we actually want: bumped both version strings, inserted a `## [1.10.0] - 2026-09-24` heading, guard passes and writes the notes. So the path this installs works on the next release, not just in principle. The `python -P` smoke test is the part I'd most want to keep — testing the artifact rather than the tree it came from is exactly the failure this catches.

The 3.12 floor is fine on our side: the reference deployment here runs **3.12.3** in all three containers, so that claim holds as stated.

Three things, one of which is ours rather than yours.

## 1. The trusted publisher has to be registered for `familiary`, not `danielplohmann`

The repository is `familiary/mcrit` now; `danielplohmann/mcrit` is a redirect. That is invisible to git clones, which is why nothing has tripped over it, but PyPI trusted publishing matches on the owner and repository GitHub reports at publish time. Registering the publisher from this PR's text or from `pyproject.toml` — both still say `danielplohmann` — would give a config that fails at the publish job with an error that does not point at the cause.

Worth updating `pyproject.toml`'s `Homepage`/`Repository`/`Issues`/`Changelog` and `RELEASING.md` to `familiary` in this PR while you are in there. The guard does not check those, so it blocks nothing; it just stops the PyPI project page pointing through a redirect.

## 2. #196-#200 need a rebase, and that is our doing

#195 went in with a rebase merge. Those five branches carry rebased copies of its content, so main now has the same changes under different SHAs and git sees both sides adding them from an older base. All five now conflict on the same four files (`CHANGELOG.md`, `docs/TUNING.md`, `mcrit/config/StorageConfig.py`, `mcrit/storage/MongoDbStorage.py`). They merged cleanly before; a merge commit would have collapsed the duplication instead.

We went with rebase because that is what you had asked for and it keeps main linear, and we would make the same call again — but the cost lands on you, so: flagging it once so you can rebase the whole tail in one pass rather than discovering it five times. **`docs/TUNING.md` wants care in that rebase** — it is mirrored verbatim into docker-mcrit and checked by CI there, so a resolution that drifts it breaks a downstream job.

This PR needs a rebase too, but only for one additive `CHANGELOG.md` hunk where both sides append under `### Added`.

## 3. The changelog gate lands on a queue that predates it

Of the 13 PRs that currently merge clean against main, only #216 and #206 touch `CHANGELOG.md`. #205 is Dependabot and exempt. The other ten do not:

**#214, #213, #212, #211, #179, #177, #175, #170, #162, #160**

Most were opened on 2026-09-06, before #188/#191 made the changelog Keep a Changelog, so this is historical rather than anything you skipped. A new workflow only fires on `pull_request` events after it merges, so they will not go red until their next push — but they will.

I would rather have the entries than the label. The release notes *are* the changelog section, so if those ten merge without entries then v1.10.0's notes are simply missing them, and this file's own header says to write the entry when the change merges rather than reconstruct it from the commit log at release time. Your PR descriptions already carry the reasoning, so it should mostly be a matter of moving it across. `no-changelog` where an entry genuinely adds nothing is fine by us.

## Where this is going

Plan is #193 first so the release runs through this rather than by hand, then the low-risk clean batch, then **v1.10.0** — minor, since everything since 1.9.0 is additive to the REST API, the config surface and the stored shape, and nothing breaks at default configuration. #196-#200 become v1.11.0 once the tail is rebased and #196's two deletion defects are fixed.

No rush on any of it from our side; say if you would rather order it differently.

danielplohmann:
Two decisions from our side, so this can move.

**Python 3.12 floor: approved.** Keep the `### Removed` entry as written. The reasoning holds here — the reference deployment runs 3.12.3 in all three containers — and one interpreter across the ecosystem is worth more than 3.11 compatibility nobody is asking for.

**The trusted publisher for `familiary` on pypi.org is already registered**, done during the org migration. So step 1 of the maintainer configuration is covered for the real index.

One gap that is worth handling in this PR, since it only shows up when the rehearsal is first run:

**test.pypi.org has no `mcrit` project at all.** That makes the TestPyPI half of the configuration a *pending* publisher rather than an ordinary one — the flow that creates the project on first publish. Without it, `workflow_dispatch` reaches the publish job and fails there, which is a confusing first experience of the rehearsal path that `RELEASING.md` recommends as the way to exercise everything safely. Worth a sentence in `RELEASING.md` saying the TestPyPI publisher has to be created as a pending publisher because the project does not exist yet.

**The published metadata still points at the old owner**, which is worth correcting while you are rebasing anyway. PyPI's `mcrit` 1.9.0 currently serves:

```
Homepage   https://github.com/danielplohmann/mcrit
Issues     https://github.com/danielplohmann/mcrit/issues
Repository https://github.com/danielplohmann/mcrit
```

Those are the `[project.urls]` in `pyproject.toml`, so the `Changelog` URL this PR adds would land pointing through the redirect too. Changing all four to `familiary` here means v1.10.0 is the first release whose project page is right.

Nothing else blocking from us: the guard, the workflow and `RELEASING.md` are good as they stand. This just needs the `CHANGELOG.md` rebase.

Separately — thank you for the #196 turnaround. I re-ran the original probe unchanged against `c4d845f` and all four cases pass, and the full suite plus ruff, ruff format and ty are clean on that head. Keeping bucket 0 until `df` reaches 0, rather than upserting it back the way I suggested, is the better fix: it repairs a hash that already lost its bucket 0 instead of only preventing new ones. #196 through #200 also merge clean against `f83a345` again.

PR #194, opened 2026-09-15
Fix two crashes that abort indexing on large corpora

r0ny123:
fork PR with CI history: https://github.com/r0ny123/mcrit/pull/43

Ran into both of these while indexing a few thousand Malpedia samples, and they're the kind that get more likely the bigger your corpus is, so worth fixing regardless of what else changes. The first is in `LogBucket.getLogBucketRange`, which raised `KeyError` for any value at or above `SHINGLER_LOGBUCKETS` — the table only covers `0..max_value-1`, and `FuzzyStatPairShingler` feeds `max_block_size`, `num_ins_C`, `num_ins_S` and `num_calls` through it without bounding any of them (only `stack_size` gets clamped, at its own call site). One basic block of 108,837 bytes was enough to kill the whole indexing job, and since the exception comes out of the hashing pool the corpus is left half-hashed. Values outside the table are clamped to its bounds now; nothing inside the table moves, so no MinHash that could be computed before changes value — the test asserts that across the whole table.

The second is `Worker.updateMinHashes` raising `UnboundLocalError` when there was nothing left to hash, because `minhashes` is only ever bound inside the batch loop. Finishing with an empty backlog therefore fails exactly like a crash, which is the normal state of a *resumed* index — that's where I hit it. While fixing it I noticed the same statement returns the size of the **last batch** rather than the total, so any run longer than one workpack silently under-reports (a 238,991-function backlog across 24 batches reported whatever the last batch happened to hold). Every caller treats it as a total — `recalculateMinHashes` assigns it to `num_updated`, `updateMinHashesForSample` substitutes 0 for a sample with no functions, `/status` hands it back as a report — so it accumulates now, which does change the number returned. Both fixes come with tests that fail against the unpatched code.


danielplohmann (approved):
Approved — verified against a deployment corpus (8,696 samples / 11.67M functions) rather than read.

**This corpus contains a live instance of the bug.** Function 5200640 has `max_block_size = 108,837`, and this deployment's LogBucket table covers `0..99999`:

| | `main` | this PR |
|---|---|---|
| `getLogBucketRange(108837)` | **RAISED KeyError** | returns 3 buckets |
| hashing function 5200640 end to end | **RAISED KeyError: 108837** | produces a signature |

108,837 is the number in your description, so this corpus holds that function or its twin. The scar is in the data too: of 11.67M functions exactly **two** are unhashed despite being hashable (`num_instructions > 10`), and 5200640 is one of them. Merging this is what lets it finally hash.

**The "no existing MinHash changes" claim holds.** 600 of the widest functions carrying stored signatures, recomputed under both trees: **0 of 600 differ**, and **600/600** recompute to exactly what is stored — so it is signature-neutral on real data.

Composed on current `main`: 331 passed, 49 subtests, ruff + format + ty clean.

One note for whoever runs `/status` after this: making `updateMinHashes` accumulate changes the number it reports, as your description says. Worth remembering when a figure looks different rather than wrong.

And the `LogBucket` cache-filename issue the docstring describes — a table built for one `max_value` being served to an instance asking for another — looks worth an issue of its own. Two instances with different `SHINGLER_LOGBUCKETS` would silently share one table, and *that* would change signatures.

PR #195, opened 2026-09-15
Add two-stage 1-vs-N matching, behind knobs that default to off

r0ny123:
Stacks on https://github.com/danielplohmann/mcrit/pull/194; only the commits after it are this PR's, and the diff will collapse to them once that one merges.

fork PR with CI history: https://github.com/r0ny123/mcrit/pull/44

1-vs-N matching is fine on a small corpus and gets steadily worse as the corpus grows, and after profiling it per stage the reason turned out to be that *every* stage scales with corpus size — and so does the answer. A query whose result names a few thousand matched samples isn't something anyone reads, and bounding the answer turns out to be the only thing that bounds the work. So this adds a cheap ranking pass in front of the exact matching: `MINHASH_MATCHING_SHORTLIST_SIZE` votes once per distinct query function (plus weighted PicHash evidence), ranks samples by vote count *and* by coverage so a small sample that matches entirely isn't buried under a large one that matches partly, and then matches only those exactly. `STORAGE_BAND_DF_CUTOFF` skips band hashes whose posting list is longer than the cutoff, decided from a stored `df` and a `(band_hash, df)` index so an over-long list is never read rather than read and thrown away — filtering on `$size` instead measured no faster than not filtering at all, since mongod reads the document to measure it. `MINHASH_PICHASH_MAX_MATCHES` bounds the PicHash path the same way, and pairwise scoring now compares each distinct MinHash signature once instead of once per function holding it, which is exact rather than approximate since a score depends only on the two signatures.

Measured on four real Malpedia corpus sizes — 2,016 / 2,997 / 5,236 / 7,244 samples, the same three query samples throughout addressed by sha256, three repeats each — the baseline's median grows as `corpus^1.40` while two-stage is `corpus^-0.07`. At the largest size that's 10.20s down to 1.58s on the median, 25.39s down to 1.93s on the tail, and 766MB down to 253MB peak RSS, with top-10 sample recall still 1.000 and 0.985 of surviving function matches keeping a bit-identical score. The memory side of this is [issue #69](https://github.com/danielplohmann/mcrit/issues/69) upstream, where workers on a 20M-function instance were reaching 60+GB. Worth being clear about the limits: 7,244 samples is a long way short of the corpus sizes this is aimed at, everything was measured on one machine with one query at a time and a warm cache, and the cutoff is a fixed constant tuned at that size while band-hash vocabulary grows sublinearly — so recall 1.000 is a measurement here, not a property of the design. All three knobs default to `0`, so nothing changes until you opt in. Stacked on https://github.com/danielplohmann/mcrit/pull/194.


danielplohmann:
Measured this on a second corpus — a byte-verified replica of a production deployment, **8,699 samples / 11,672,873 functions**, also Malpedia. Production was never written to; the replica was taken with the stack stopped and confirmed functionally identical to live (same query, same result digest).

**The quality claim holds exactly, and that is the headline.** Both prerequisites were built first (`rebuildFunctionRangeIndex`: 8,405 samples in 65 s; `rebuildBandDfIndex`: 7,208,816 documents in 190 s), then `MINHASH_MATCHING_SHORTLIST_SIZE` 0 → 100 as the only variable, ABBA ordering.

| query | functions | baseline | shortlist=100 | speedup | matched samples | top-10 | top-25 |
|---|---|---|---|---|---|---|---|
| win.citadel | 850 | 52.4 s | 24.3 s | **2.15x** | 5,524 → 3,347 | **1.000** | **1.000** |
| win.merlin | 7,018 | 466.5 s | 280.2 s | **1.66x** | 3,287 → **178** | **1.000** | **1.000** |

Medians, 4 and 3 runs per arm respectively. **Top-25 recall is 1.000 on both**, across very different truncation — merlin keeps 178 of 3,287 samples and still loses nothing a reader would look at. That is your central claim, confirmed on a corpus you could not test.

## The speedup is lower here, and it falls as the query grows

You report **6.4x on the median** at 7,244 samples. This corpus gives **2.15x** and **1.66x**. Since both are Malpedia, the dataset is not the variable. Two things that are:

**1. The query size.** Your three query samples are win.acidbox (158 fn), win.zloader (557 fn) and win.blackpos (650 fn) — all under 700. Merlin is **10.8x larger than any of them**, and it gets the *smaller* speedup.

That looks mechanistic rather than noisy: the shortlist bounds how many **corpus samples** the exact stage examines, but a 7,018-function query still has to band-query, fetch and score **all of its own** functions, and none of that is bounded. So the bounded share of total cost shrinks as the query grows. If that reading is right, the factor a deployment sees depends on its *query* sizes as much as its corpus size — and the suite currently has no large-query point to show that.

**2. The band projection.** This deployment does not run stock banding: it uses an explicit 15-band `segment-balanced-v1` projection covering 64/64 signature fields (stock random projection reached 45). The shortlist votes on band posting lists, so a different projection changes what stage 1 sees. Related, and measured on the same corpus for #196: the largest band document here holds **80,973 postings / 879,653 bytes** — 10.86 bytes each, which reproduces your ~10.4 almost exactly — but that is **19.1x headroom to the 16MB cap, not 84.9x**, putting the wall nearer 166,000 samples than 615,000. That *strengthens* your "the hard limit is nearer than sharding" conclusion by about 4x.

## Caveats on my numbers

- **The seconds are not production numbers.** The replica runs a 2 GB WiredTiger cache against the live instance's ~15 GB default, deliberately — two default-cache mongods do not fit this box. The **ratios** are the finding.
- **Two query samples, one k (k=1), one shortlist size (100).** `STORAGE_BAND_DF_CUTOFF` and `MINHASH_PICHASH_MAX_MATCHES` stayed off, so this measures the shortlist alone rather than the full package.
- 3,347 and 178 both exceed 100 because PicHash matches are bounded by a separate knob that was off — not a bug, just worth stating.
- Also worth noting for anyone reproducing: two runs per arm gave 2.87x on citadel; four gave 2.15x. The first shortlist run warms the cache for the second, and a thin sample lets that outlier set the answer.

None of this is a refutation — the direction is right, the quality claim is exact, and the mechanism is sound. It is a second point suggesting the factor is corpus- *and* query-dependent, which seems worth knowing before someone plans capacity around 6.4x.

Happy to run more query sizes, other shortlist values, or the DF cutoff on this corpus if any of that would be useful to you.

r0ny123:
Thank you, this is the measurement the PR was missing. Your reading of merlin fits the code: the shortlist bounds the corpus side of stage 2, but band lookup, posting fetch and candidate scoring for the query's own functions stay proportional to the query size. So a 7,018-function query keeps most of its cost, and my three sub-700-function queries were the best case. I've added a paragraph to the description saying the speedup depends on query size as well as corpus size, citing your numbers, so nobody plans capacity around 6.4x.

The band projection point is useful too. Stage 1 votes on whatever the posting lists contain, so a 15-band `segment-balanced-v1` projection is a different input from stock banding. Recall held at 1.000 under it, which is reassuring.

If you still have time on the replica, the most useful run would be merlin again with `STORAGE_BAND_DF_CUTOFF` on as well as the shortlist. The cutoff is the knob aimed at the query-side cost the shortlist leaves alone, so it's the natural thing to check next. A mid-sized query of around 2-3k functions would also help show where the speedup falls between 850 and 7,018.

danielplohmann (dismissed):
Approved, measured rather than read. Defaults-off is byte-identical on a real 1-vs-N query against 8,699 samples / 11.67M functions (same digest, same 6,809,951 pairs, same 607 MiB fetched), so this changes nothing for a deployment that does not opt in. With the shortlist enabled on a replica: top-25 recall 1.000 on both query samples tested, at 2.15x and 1.66x. Full numbers and caveats in the comment above. The fail-open-with-a-warning behaviour when the prerequisites are missing is the right call and I verified it fires.

danielplohmann:
Ran both. **The DF cutoff is the missing half, and it changes the conclusion I drew from the shortlist alone.**

Same replica (8,699 samples / 11,672,873 functions, Malpedia, 15-band `segment-balanced-v1`), k=1, `MINHASH_MATCHING_SHORTLIST_SIZE=100`, `STORAGE_BAND_DF_CUTOFF=200` — your measured starting point. Medians of three runs per arm.

| query | functions | baseline | + shortlist | + shortlist & DF | SL | **SL+DF** |
|---|---|---|---|---|---|---|
| win.citadel 601 | 850 | 52.4 s | 24.3 s | **7.8 s** | 2.15x | **6.67x** |
| win.industroyer 3665 | 2,695 | 121.9 s | 43.7 s | **22.5 s** | 2.79x | **5.43x** |
| win.merlin 5188 | 7,018 | 466.5 s | 280.2 s | **21.2 s** | 1.66x | **22.03x** |

**Top-10 and top-25 recall are 1.000 in every arm measured**, including both SL+DF configurations. Matched-sample counts under SL+DF: 3,346 / 3,868 / 168.

## My "the speedup falls as the query grows" was an artifact of the partial configuration

You were right to ask for the cutoff. With the shortlist alone the factor did fall with query size (2.15x → 2.79x → 1.66x, worst on the largest query). **With both knobs the picture is the opposite of what I reported**, and better than either of us framed it:

```
baseline :  52.4s → 121.9s → 466.5s     (scales with query size, ~9x across the range)
SL + DF  :   7.8s →  22.5s →  21.2s     (essentially flat above ~2.7k functions)
```

The two knobs together **flatten the cost curve** rather than scaling it down. A 7,018-function query and a 2,695-function one now cost within 6% of each other, where at baseline the larger cost 3.8x more. So the speedup is large precisely where the baseline was expensive — 22x on merlin — and the right way to state it is not a factor at all but *query time stops tracking query size*, which is what "bounding the answer is the only thing that bounds the work" predicts.

That also explains the shortlist-only result cleanly: the shortlist bounds stage 2's corpus side, the cutoff bounds the query side, and measuring one without the other on a large query shows the unbounded half dominating. My earlier comment should be read with that correction — happy for you to quote this one instead.

## Caveats unchanged

- Replica runs a 2 GB WiredTiger cache against the live instance's ~15 GB default, so the **ratios** are the finding, not the seconds.
- Three query samples, one k (k=1), one shortlist size (100), one cutoff (200). No sweep of either knob.
- `MINHASH_PICHASH_MAX_MATCHES` stayed off throughout.
- The corpus is Malpedia but a different snapshot from yours, and the projection is non-stock.

Glad to sweep the cutoff (e.g. 100/200/1000) or add larger queries if that would help pin the flat region.

r0ny123:
Thank you, that settles it. I've replaced the query-size caveat in the description with your three-query table and the "query time stops tracking query size" framing, and it now says to enable the two knobs together. It links this comment rather than the earlier one.

The corpus side and query side each bounding one half matches how the stages are built, so I don't think the PR needs a cutoff sweep before merging. A sweep would be worth doing as a follow-up to choose a recommended default. The 200 was tuned at 7,244 samples, and band vocabulary grows sublinearly, so the right value probably drifts with corpus size.

danielplohmann (dismissed):
Re-approving. The measurement you asked for is in (#issuecomment-5816650497) and it settled the question — the two knobs bound different sides and together the cost stops tracking query size, which is better than either of us had it before.

Verified the rebase onto current main (`ab56c34`, i.e. after #194) locally before this: three commits replay cleanly, the indexing-crash-fixes commit drops as already applied, and the full suite plus ruff, ruff format and ty pass on the rebased tree against a throwaway mongod. Merging with rebase.

PR #196, opened 2026-09-15
Let a band posting list outgrow a single MongoDB document

r0ny123:
Stacks on https://github.com/danielplohmann/mcrit/pull/195; only the commits after it are this PR's, and the diff will collapse to them once that one merges.

fork PR with CI history: https://github.com/r0ny123/mcrit/pull/45

A band posting list lives as a `function_ids` array inside one document, and MongoDB caps a document at 16MB, which puts a hard ceiling on corpus size that I don't think anyone has hit yet but is closer than it looks. On a 7,244-sample corpus the largest `band_0` document held 18,968 postings in 197,606 bytes — about 10.4 bytes each, so roughly 1.6M fit, which is 84.9x that corpus and puts the wall somewhere near 615,000 samples. I checked rather than extrapolating: pushing 100,000 ids at a time into one document succeeded ten times and failed on the eleventh with `BSONObj size: 17588958 is invalid`. It *fails*, it doesn't slow down, so indexing stops outright for any sample holding a function whose band hash is already at the cap. Sharding doesn't help either, since a document can't span shards.

`STORAGE_BAND_BUCKET_SIZE` splits a hash across `(band_hash, bucket)` documents once it would exceed the cap, and defaults to `0`, which keeps the current single-document shape byte for byte. Bucket 0 carries the bookkeeping for the whole hash — `df` as the total across every bucket, plus `tail`/`tail_n` for placement — which is what leaves the cutoff filter and its `(band_hash, df)` index untouched: a hash under the cutoff is nowhere near a bucket's worth so it never spills, and one that did spill has a `df` that rejects it anyway. Buckets fill in order rather than by hashing the function id, deliberately, since modulo placement would scatter a 200-entry posting list across 200 documents and make the common cheap case expensive. One thing to watch on an existing database: enabling this needs `rebuild_band_df_index` run first, because documents written earlier have no `bucket` field and the upsert filter would miss them and insert a *second* document per hash, splitting the list with nothing reporting an error — the rebuild stamps `bucket: 0` and is what makes them addressable. The estimate above is also on the optimistic side, since Malpedia is curated and deduplicated; a corpus full of near-duplicate packed variants would concentrate `df` faster and get there sooner. Stacked on https://github.com/danielplohmann/mcrit/pull/195.


danielplohmann:
Reviewed this one properly for the first time. The problem is real and the design reads well — bucket 0 carrying the whole hash's bookkeeping so the `(band_hash, df)` cutoff index stays untouched is a nice result, and filling in order rather than by modulo is clearly right for the short-posting-list common case. Correcting your own 615k estimate down to 270k in 57079ee rather than leaving the optimistic number standing is the part I'd most want to keep.

Two defects in the **deletion** path though, and both are silent. I demonstrated them on a throwaway mongod rather than reasoning from the code, probe at the bottom.

## 1. A bucketed pull only ever reaches one document

The pull branch in `_updateBands` is unchanged by this PR:

```python
band_updates.append(UpdateOne({"band_hash": band_hash}, {"$pull": {"function_ids": {"$in": function_ids}}}))
```

`UpdateOne` matches a single document. That was exactly right when one hash was one document. Under bucketing a spilled hash is N documents, and the pull reaches whichever one matches first — in practice bucket 0. Postings in buckets >= 1 are never removed.

With `STORAGE_BAND_BUCKET_SIZE=2`, one hash, 7 postings (buckets 0..3):

```
victims [5,6,7]  (buckets 2 and 3)
  unbucketed  ->  df=4  surviving=[1,2,3,4]           correct
  bucketed    ->  df=7  surviving=[1,2,3,4,5,6,7]     nothing removed at all

victims [1,5]    (bucket 0 and bucket 2)
  bucketed    ->  df=6  surviving=[2,3,4,5,6,7]       id 5 survives
```

Reachable from `deleteSample`. `_recomputeBandBookkeeping` then recomputes `df` from what remains, so the counters end up perfectly consistent with the *stale* content — which is why `testDeletionKeepsBookkeepingConsistent` passes. It asserts `df == sum of surviving postings`, and never that the deleted sample's postings are gone. That assertion holds no matter how much leaks.

How much it matters depends on the other knob. With `STORAGE_BAND_DF_CUTOFF` on and well under the bucket size, a spilled hash is rejected at lookup anyway and the leaked postings are never read. With the cutoff **off, which is its default**, `_bandLookupPipeline` returns `{"band_hash": {"$in": ...}}` and reads every bucket, so they go straight into candidate accumulation and the matcher scores against deleted functions.

`UpdateOne` -> `UpdateMany` looks like it covers it; it doesn't upsert, so the #149 "a pull must not create a document" property survives.

## 2. Bucket 0 can be deleted, taking the hash's bookkeeping with it

The pull is followed by `delete_many({..., **_EMPTY_BAND_DOCUMENT})`, which removes any bucket the pull emptied — bucket 0 included. But bucket 0 is the only document holding `df`/`tail`/`tail_n` for the whole hash, and `_recomputeBandBookkeeping` writes with `UpdateOne(..., {"$set": ...})` and no `upsert`, so it can't bring it back.

Same setup, victims = all of bucket 0:

```
after = [(1, None, None, None, [3,4]), (2, None, None, None, [5,6]), (3, None, None, None, [7])]
```

Five live postings and no `df` anywhere. Two ways that bites:

- with the cutoff on and the df index complete, `{"df": {"$lte": cutoff}}` is type-bracketed and doesn't match a missing field, so the **whole hash disappears from lookups** — five postings silently dropped from matching;
- the next push reads state through `{"bucket": {"$in": [0, None]}}`, finds nothing, restarts at `tail=0, tail_n=0`, creates a fresh bucket 0 and sets `df` to just the new batch, undercounting by five.

The docstring anticipates a gap in the middle of the chain ("bucket 2 surviving while bucket 1 emptied") but not bucket 0 itself going. Either excluding bucket 0 from the empty-document delete while another bucket for that hash survives, or `upsert=True` on the bookkeeping restore, would do it.

## Three smaller things

- **`STORAGE_BAND_DF_CUTOFF > STORAGE_BAND_BUCKET_SIZE` truncates silently.** The fast path matches `{"band_hash": ..., "df": {"$lte": cutoff}}`, and only bucket 0 has `df`, so a spilled hash whose total `df` still fits under the cutoff returns bucket 0 alone — a truncated posting list, no error. The StorageConfig comment does state the constraint, so this is documented rather than unknown; it's just prose guarding a silent wrong answer, and an assertion at startup would be cheap.
- **Same document written twice in one unordered bulk.** In `_pushBandsBucketed`, when a batch starts in bucket 0 and spills upward, the bucket-0 write carries `$set: {"tail": 0, "tail_n": ...}` and a later `UpdateOne` in the same `bulk_write(..., ordered=False)` corrects those to the final values. Unordered bulks have no defined execution order. mongod processes the `updates` array in order today so it isn't broken, but dropping `tail`/`tail_n` from the bucket-0 push and letting the trailing update own them unconditionally would make the two writes touch disjoint fields and stop it mattering.
- **Dead branch.** `elif method == "push_bucketed": pass` is unreachable — `method` is validated to be `push` or `pull` at the top, and the bucketed path returns before that loop.

## Not defects, just noting

The migration requirement is documented clearly and `_rebuildBandBookkeepingBucketed` is written to double as it, but nothing enforces the ordering. A `$setOnInsert: {"bucket": 0}` with a `{"bucket": {"$in": [0, None]}}` filter on the upsert would make it self-migrating and remove the requirement. Separately, the read-tail-then-place sequence is a TOCTOU between concurrent workers, so `tail_n` can undercount and a bucket overshoot `bucket_size` — with 100,000 against a measured ~1.05M capacity that's tidiness rather than correctness, and the non-unique `(band_hash, bucket)` index can't prevent two concurrent upserts both missing either.

## Probe

Drop in as `tests/testProbeBucketedPull.py` and run with `-m mongo`:

```python
import logging, unittest
import pytest
from mcrit.config.McritConfig import McritConfig
from mcrit.config.StorageConfig import StorageConfig
from mcrit.storage.MongoDbStorage import MongoDbStorage
from mcrit.storage.StorageFactory import StorageFactory
from .context import getTestMongoServerAndPort

logging.disable(logging.CRITICAL)


@pytest.mark.mongo
class ProbeBucketPull(unittest.TestCase):
    def _storage(self, bucket_size, suffix):
        server, port = getTestMongoServerAndPort()
        c = McritConfig()
        c.STORAGE_CONFIG = StorageConfig(
            STORAGE_METHOD=StorageFactory.STORAGE_METHOD_MONGODB,
            STORAGE_SERVER=server, STORAGE_PORT=port,
            STORAGE_MONGODB_DBNAME="probe_pull_" + suffix,
            STORAGE_BAND_BUCKET_SIZE=bucket_size)
        st = MongoDbStorage(c)
        st.clearStorage()
        return st

    def _dump(self, st):
        return sorted(
            (d.get("bucket"), d.get("df"), d.get("tail"), d.get("tail_n"), list(d.get("function_ids") or []))
            for d in st._getDb()["band_0"].find({"band_hash": 4242}))

    def _run(self, bucket_size, suffix, victims):
        st = self._storage(bucket_size, suffix)
        st._updateBands({0: {4242: list(range(1, 8))}}, method="push")
        st._updateBands({0: {4242: victims}}, method="pull")
        after = self._dump(st)
        return after, sorted(i for _b, _d, _t, _n, f in after for i in f)

    def testPullUnbucketed(self):
        _after, surviving = self._run(0, "unbucketed", [5, 6, 7])
        self.assertEqual(surviving, [1, 2, 3, 4])

    def testPullBucketed(self):
        _after, surviving = self._run(2, "bucketed", [5, 6, 7])
        self.assertEqual(surviving, [1, 2, 3, 4])

    def testPullBucketedMixed(self):
        _after, surviving = self._run(2, "mixed", [1, 5])
        self.assertEqual(surviving, [2, 3, 4, 6, 7])

    def testBucketZeroEmptied(self):
        after, _surviving = self._run(2, "zero", [1, 2])
        self.assertTrue(any(d[0] == 0 for d in after), "bucket 0 must survive to carry df/tail/tail_n")
```

`testPullUnbucketed` passes, the other three fail.

Both fixes look small and they're your branch, so I'd rather hand them over than push into it. #195 I'll take now — it's independent and you've said it's ready. #197 stacks on this one so it waits on the same round; nothing in its own content is implicated.

r0ny123:
Thanks for testing this against a real mongod and including the probe. Both deletion defects reproduced exactly as you described, and everything is fixed in fb63d5d. I wrote the tests first from your reproductions (bucket size 2, one hash, 7 postings; victims [5,6,7], [1,5] and all of bucket 0) and confirmed they failed on the previous head.

1. A pull only reached one bucket. It now uses `UpdateMany` with no upsert, so a pull against an unknown hash still doesn't create a document (#149). There's a test for that too. `testDeletionKeepsBookkeepingConsistent` now also checks that none of the deleted sample's function ids are left in any band collection, and the new tests check the exact surviving postings and `df` for both of your victim sets.

2. Bucket 0 could be deleted. I went with keeping bucket 0 rather than only upserting it. Under bucketing a pull now deletes the emptied buckets above 0 first, then recomputes the bookkeeping, and removes bucket 0 only once `df` reaches 0. At that point no bucket of the hash holds a posting, and a test checks that pulling everything leaves nothing behind. The recompute also upserts bucket 0 while `df > 0`, which repairs a hash that already lost its bucket 0. `purgeEmptyBandDocuments` would otherwise have removed an empty bucket 0 that still carries a nonzero `df`, so it now skips those. The bucket-0 test also pushes again afterwards and asserts that `df` is 7, that placement continues from the real tail, and that no bucket goes over the size.

On the smaller items:
- Cutoff above bucket size: `MongoDbStorage` now raises `ValueError` at construction when `STORAGE_BAND_DF_CUTOFF > STORAGE_BAND_BUCKET_SIZE` and both are set. There's a test for it, and the StorageConfig comment says so.
- Same document written twice in one unordered bulk: the bucket-0 `$push` no longer sets `tail`/`tail_n`. The trailing write now always owns them, so the two writes touch different fields. That trailing write upserts only when no bucket-0 push comes before it in the batch, so one batch never has two upserts for the same new document.
- Dead branch: removed.

I've left the migration ordering and the concurrent-worker TOCTOU as they are for now, since they're design choices rather than bugs. The self-migrating `$setOnInsert` with the `$in: [0, None]` filter is a good idea, and I'd rather do it as its own follow-up with a test against a pre-bucketing collection than fold it into this fix. On the TOCTOU I agree with your read: with 100,000 against about 1.05M it only affects how tidy the buckets are, not correctness.

ruff, ruff format, ty and the full suite pass on fb63d5d: 341 passed, including the mongo-marked tests against a local mongod 7.0. The 8 sleep-marked tests were not run.

